PR #195, opened 2025-11-05
docs(test): validation report reproducibility + regex guard test

weissarc:
- Add reproducibility block (rules/dataset commit, commands) to VALIDATION_REPORT.md
- Add regex guard unit test to prevent YAML escaping regressions for \s patterns

Non-functional changes; safe to merge after CI.


cyprobot (changes requested):
Recording the verdict from my review above as a review rather than a comment, so it is attached to `7b8cfafb` and goes stale if the branch moves.

Two changes before this is a yes:

1. **The guard must read the shipped rules.** `TestRegexWhitespaceGuard_HTTPServerHeader` builds its own rule from a Go string literal, so it asserts that Go's escaping works — not that the YAML's does. Copied onto current `main`, where `imap.generic` carries exactly the defect this test is named for (#241), it passes. Point it at `loadBuiltinRules()` and it fails today, which is how you know it measures.

2. **Re-pin the reproducibility block.** `Rules commit: 27e5d35` is a 27-rule database; `main` is 50 rules at `16a5b56`. A stale pin is worse than none.

Detail and the reproduction are in the comment above. The reproducibility block itself is worth having — this is about the shape it lands in, not whether.

btuvay:
Reviewed at `7b8cfaf` against current `main` (`d410c71`). CI is green and the test still compiles and passes on today's tree, so nothing here is stale in the mechanical sense. Two things worth fixing before this lands, one of them substantive.

**The guard does not exercise the thing it guards.** The test builds its rule as a Go literal:

```go
Match: "server:\\s*nginx",
```

so no YAML is parsed anywhere in it. A YAML escaping regression — the failure `#194` fixed — cannot fail this test, because the string never passes through the loader. What it actually asserts is that Go's `regexp` handles `\s*`, which the standard library guarantees. If someone reintroduced the original defect tomorrow, this test would stay green.

**What would catch it**, and is barely longer: assert on the rules after they come out of the loader.

```go
rules := loadBuiltinRules()
for _, r := range rules {
    if strings.Contains(r.Match, `\\`) { /* escaping was doubled: the pattern can never match */ }
    if _, err := regexp.Compile(r.Match); err != nil { /* ... */ }
}
```

The regression is silent in exactly one direction: a pattern written `\\s` inside a single-quoted YAML scalar loads as a literal backslash followed by `s`, compiles fine, and never matches anything. Compilation is not the gate — the doubled backslash is.

Measured on `d410c71` so the numbers are in the thread: 52 rules load, 21 of them carry `\s`, none carries a literal `\\`, and all 52 compile. The file mixes both styles deliberately (26 pattern lines use `\\s` inside double quotes, 24 use `\s` in single quotes), which is correct in both cases and is exactly why a loader-level assertion is worth having — the two styles are one editing mistake apart.

**The reproducibility block pins a commit that no longer reproduces.** `27e5d35` is the head of `#194`, 2025-11-05. Rules and dataset have both moved a long way since, so as written the block claims a correspondence that is no longer true. Either regenerate the three values at merge time, or phrase it as "as of `27e5d35`" so a reader knows it is a historical record rather than a current one. `scripts/compare-validation-metrics.sh` does still exist, so the commands themselves are fine.

Not blocking from our side — we do not hold merge rights here. The doc half is harmless once the wording is fixed; the test half is the one that would otherwise sit in the suite looking like coverage it does not provide.

PR #246, opened 2026-08-13
chore: build with Go 1.26.6 to close three reachable stdlib vulnerabilities

### LINKED ISSUE #245: Three of four reachable vulnerabilities are stdlib and no dependency bump closes them: CE is pinned to Go 1.25
## Summary

Three of the four reachable vulnerabilities in this repository are in the Go standard library, not in any module we depend on. No dependency bump closes them — only a toolchain upgrade does. We are pinned to Go 1.25 in both `go.mod` and CI.

Surfaced while reviewing #216 (`golang.org/x/crypto` bump). That PR is a fair example of the confusion: it upgrades six `x/` modules and closes none of these.

## Reachable set

| id | package | closed by |
|---|---|---|
| GO-2026-5856 | `crypto/tls` | toolchain |
| GO-2026-5039 | `net/textproto` | toolchain |
| GO-2026-5037 | `crypto/x509` | toolchain |
| GO-2026-5942 | `golang.org/x/net` | `x/net` ≥ 0.56.0 |

The first three require Go 1.26.4 / 1.26.5. The fourth is a module bump and is being handled on #216.

## Current pinning

```
go.mod                          go 1.25.0
.github/workflows/test.yaml     GO_VERSION: '1.25'   (and matrix go-version: ['1.25'])
.github/workflows/validate.yaml GO_VERSION: '1.25'
.github/workflows/codeql.yml    go-version-file: go.mod
```

`codeql.yml` reads the version from `go.mod`, so raising `go.mod` moves it too. The other two are pinned separately and need editing as well, or they will keep building on 1.25 after `go.mod` changes.

## Why this matters more than the dependency bumps

`crypto/tls` and `crypto/x509` are on the path this product exercises constantly — every TLS probe, every certificate read, every ssl-template evaluation. `net/textproto` is on the HTTP path. These are not transitive libraries pulled in by something else; they are the standard library under the scanner's main workload.

By contrast, `golang.org/x/crypto` is `// indirect` here, pulled in by `github.com/go-playground/validator/v10` for `sha3`, with zero direct imports anywhere in the tree.

## Work

1. Raise `go` in `go.mod` to the target toolchain.
2. Raise `GO_VERSION` in `test.yaml` and `validate.yaml`, and the `go-version` matrix entry.
3. Re-run `govulncheck` and confirm the three stdlib entries drop out.
4. Check the appliance and release pipelines for a separately pinned Go version — this issue covers CE's own workflows only.

## Provenance

The `govulncheck` comparison (main versus #216 head, identical reachable set) was measured by the session working on #216, not in this issue. The version pinning above is verified directly from the files. Anyone acting on this should re-run `govulncheck` themselves rather than trust the table — it is a snapshot, and the database moves.

cyprobot:
Closes #245.

## What this closes, measured

`govulncheck ./...` on this branch, before and after:

```
before (go1.26.3)                 after (go1.26.6)
GO-2026-5856  crypto/tls          No vulnerabilities found.
GO-2026-5039  net/textproto       0 reachable
GO-2026-5037  crypto/x509
3 reachable
```

All three are standard library, so no dependency bump reaches them — #216 upgraded six `x/` modules and closed only the fourth one, `GO-2026-5942` in `x/net`.

The reachability is not theoretical. `crypto/tls` is reached from six call sites including `banner_grab.go` and `tls_native_probe.go`, `net/textproto` from `http_parser.go:277`, `crypto/x509` from TLS certificate verification — the paths a scanner runs against every host it touches.

## Why the go directive is not raised

`go 1.25.0` stays. What compiles the standard library into the binary is **the toolchain**, not the language version, so `toolchain go1.26.6` is what closes these. Raising the `go` directive would additionally force every downstream consumer — vulntor-ee imports this module — onto 1.26 language semantics while closing nothing extra.

With `GOTOOLCHAIN=auto` (the default), a developer on an older Go downloads 1.26.6 automatically; this was observed rather than assumed:

```
$ go version
go: downloading go1.26.6 (darwin/arm64)
go version go1.26.6 darwin/arm64
```

## This does not close the vulnerability in the shipped product

A `toolchain` directive travels with the module but **does not propagate to consumers** — only the `go` line does. vulntor-ee imports this module and needs the same change in its own `go.mod`. That sentence is easy to read as a footnote. Measured, it is not one.

EE builds what the customer runs, and every one of its build paths is on 1.25:

```
EE go.mod                          go 1.25.0, no toolchain directive
EE ci.yml / pr-review-gate.yml     GO_VERSION: '1.25'
EE release.yaml                    go-version: '1.25'      <- ships binaries
EE release-staging.yaml            go-version: "1.25"      <- ships binaries
EE _build-ee-binary.yml            default: '1.25', and no caller overrides it
```

The same CE tree, resolved against the line EE actually builds with:

```
go1.25.8   ->  8 vulnerabilities
go1.26.3   ->  3
go1.26.6   ->  0
```

So when this merges, CE is clean and the shipped enterprise binary still carries **eight** — more than CE had before this PR, not fewer. Nothing in the diff is wrong; the risk is that #245 reads as closed on the strength of its first sentence.

**EE's fix is smaller than this one.** All eight close inside the 1.25 line:

| id | fixed in |
|---|---|
| GO-2026-4870, 4946, 4947 | 1.25.9 |
| GO-2026-4918, 4971 | 1.25.10 |
| GO-2026-5037, 5039 | 1.25.11 |
| GO-2026-5856 | 1.25.12 |

Pinning EE to `1.25.12` closes all eight: no language version change, no effect on consumers, and none of the `go`-versus-`toolchain` argument above applies there. Verified by resolving this tree at 1.25.12 — zero reachable. Filed separately so it is not lost.

## A fourth pin the issue did not list

#245 named `go.mod`, `test.yaml` and `validate.yaml`. There is one more:

```
.github/workflows/validate-fingerprints.yaml:22   go-version: '1.24'
```

Two minor versions behind, and it runs the fingerprint validation suite. Moved with the rest. `codeql.yml` reads `go-version-file: 'go.mod'` and needs no edit.

## Verification under 1.26.6

| check | result |
|---|---|
| `go build ./...` | passes |
| `go test ./...` | passes, no failures |
| `golangci-lint run ./...` (v2.12.2, the CI-pinned version) | exit 0, 0 issues |
| `govulncheck ./...` | 0 reachable |

The linter was worth checking rather than assuming: a golangci-lint binary built with an older Go cannot read newer export data. The prebuilt release CI downloads reports `built with go1.26.2`, same major, so it reads 1.26.6 fine — confirmed by running that exact binary against this branch, not the one `go install` would build locally.

`go vet` reports eight `lostcancel` findings in `mysql_native_probe.go`, `smtp_native_probe.go` and `winrm_native_probe.go`. These are **pre-existing**: the identical eight appear under 1.25.0. Not touched here, and worth their own issue.

## What still is not measured

Nothing in CI runs `govulncheck`, so the next reachable stdlib vulnerability will land the same way this one did — noticed by hand while reviewing an unrelated dependency PR. A CI job would turn that into a build signal, but it is a separate change and is deliberately not bundled here.

## Left alone deliberately

`.golangci.yaml:491` still declares `go: '1.24'`. That is the language version the linter assumes, not what compiles the binary, so it carries no security weight — but it is stale in the same way and changing it can change lint strictness, which does not belong in a vulnerability fix.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -10,7 +10,7 @@ on:
       - '*'
 
 env:
-  GO_VERSION: '1.25'
+  GO_VERSION: '1.26.6'
   NODE_VERSION: '20'
   VERSION: 'dev'
 
@@ -118,7 +118,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
-        go-version: ['1.25']
+        go-version: ['1.26.6']
 
     steps:
       - name: Check out code
--- a/.github/workflows/validate-fingerprints.yaml
+++ b/.github/workflows/validate-fingerprints.yaml
@@ -19,7 +19,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
-          go-version: '1.24'
+          go-version: '1.26.6'
           check-latest: true
 
       - name: Run validation suite (PR)
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -6,7 +6,7 @@ on:
       - '*'
 
 env:
-  GO_VERSION: '1.25'
+  GO_VERSION: '1.26.6'
   GOLANGCI_LINT_VERSION: v2.12.2
   MISSPELL_VERSION: v0.6.0
 
--- a/go.mod
+++ b/go.mod
@@ -2,6 +2,8 @@ module github.com/cyprob/cyprob
 
 go 1.25.0
 
+toolchain go1.26.6
+
 require (
 	github.com/Masterminds/semver/v3 v3.4.0
 	github.com/charmbracelet/lipgloss v1.1.0


cyprobot (commented):
Verified the vulnerability claim by running it rather than reading it, on this tree at `8247bae`:

```
GOTOOLCHAIN=local (go1.26.3)
  GO-2026-5856  crypto/tls@go1.26.3      fixed in 1.26.5
  GO-2026-5039  net/textproto@go1.26.3   fixed in 1.26.4
  GO-2026-5037  crypto/x509@go1.26.3     fixed in 1.26.4
  -> affected by 3 vulnerabilities from the Go standard library
```

Three, all stdlib, highest fix at 1.26.5 — so 1.26.6 closes them with one patch to spare. Your measurement holds.

The pin sweep also holds. All five workflows carry a Go setup or none: `test.yaml`, `validate.yaml` and `validate-fingerprints.yaml` are on `1.26.6`, `sync-docs.yml` has no Go, and `codeql.yml` reads `go-version-file: go.mod`. The fourth pin you found was real and the issue did miss it.

## The `toolchain` choice is right, and for a better reason than the PR gives

The PR argues it on cost — not forcing 1.26 language semantics on consumers. True, but the stronger argument is reach: `toolchain` travels **with the module**, so `make build` picks up 1.26.6 for anyone, not only for CI. Since this repo has no release workflow and no goreleaser — `make build` is plain `go build` — pinning only the CI versions would have left every local and out-of-band build on whatever the builder happened to have. The directive covers that; the CI pins alone would not.

I checked the one thing that defeats it: `GOTOOLCHAIN` is not set anywhere in either repo, so nothing overrides the directive back down.

## Where I would change the PR: the stated cost understates it, by a factor

The body says EE does not benefit. That is true and it reads like a footnote. Measured, it is the opposite of a footnote:

```
EE go.mod        go 1.25.0    (no toolchain directive)
EE CI pins       1.25         (every workflow)

same CE tree, built with the line EE actually uses:
  GOTOOLCHAIN=go1.25.8  ->  affected by 8 vulnerabilities
  GOTOOLCHAIN=go1.26.3  ->  affected by 3
  GOTOOLCHAIN=go1.26.6  ->  0
```

The product customers run is built by EE. So after this merges, the CE repo is clean and **the shipped enterprise binary still carries eight** — more than CE had before this PR, not fewer. Nothing here is wrong, but "EE does not benefit" and "the exposure this issue is about is untouched where it ships, and is 2.7× larger there" are different sentences, and #245 will read as closed on the strength of the first one.

## And EE's remedy is smaller than this one, which is worth saying explicitly

Every one of the eight is fixed inside the 1.25 line:

```
GO-2026-4870, 4946, 4947  ->  1.25.9
GO-2026-4971, 4918        ->  1.25.10
GO-2026-5037, 5039        ->  1.25.11
GO-2026-5856              ->  1.25.12
```

So EE closes all eight by pinning `1.25.12` — no language bump, no consumer impact, and none of the `go`-versus-`toolchain` argument applies to it at all. Somebody reading this PR would reasonably assume EE needs the same 1.26 move it just debated. It does not, and saying so in the follow-up saves that debate from being had twice.

## Agreed, no argument

- **`.golangci.yaml:491` left alone.** It is the linter's assumed language version, it does not build anything, and changing it changes lint strictness for no security gain. Correct to keep out of a vulnerability fix.
- **The two things you refused to assume.** Downloading the prebuilt golangci-lint binary rather than `go install`-ing it locally is the right call and the reason matters: a locally built linter would have been compiled by your own 1.26.6 and told you nothing about whether CI's binary can read 1.26 export data. That is the same trap as a harness in the wrong shell.
- **`go vet` lostcancel ×8.** Confirmed pre-existing rather than introduced is the right check; leaving them is right.

## On `govulncheck` in CI

File it. Keeping it out of a vulnerability fix is correct, and there is now a concrete argument for the follow-up: the 8-versus-3 gap between what CE builds and what EE ships was invisible until somebody ran the tool by hand today, on a PR about something else. A check that runs on both repos would have surfaced it as a number rather than as a review comment.

## Verdict

Correct and complete for this repo. The two things I would change are in the description, not the diff: state EE's exposure as measured rather than as "does not benefit", and name `1.25.12` as EE's actual remedy so the follow-up does not re-litigate this PR's decision.

Merge is not mine. Integration Tests and the macOS build were still running when I looked.

Issue #239, opened 2026-08-13
A candidate dropped by the 0.50 confidence threshold is indistinguishable from one that never matched


cyprobot:
> **Rewritten 2026-08-13.** The original title and body claimed that an error page can make a device unfingerprintable. That claim did not survive measurement and the original framing is preserved in the comments below, along with how it was narrowed. The defect that remains is real but smaller and different in kind.

## Summary

A candidate removed by the `conf < 0.50` threshold is indistinguishable, downstream, from one that never matched. Both produce an absent identification and nothing in the output says which happened.

This matters independently of how often the threshold actually bites — which, as measured below, is currently **never**, on anything anyone has captured.

## The mechanism

`pkg/fingerprint/resolver_rulebased.go`:

```go
softPenalty := softExcludePenalty(normalizedBanner, rule.softExRegex, 0.20)
score := base - softPenalty + portBonus
conf  := calculateConfidence(base, softPenalty, portBonus)

if conf < 0.50 {
    continue          // removed from the candidate list, not demoted
}
```

`softExcludePenalty` charges 0.20 **per matching pattern**, not per occurrence — a page repeating `forbidden` four times pays 0.20 once.

## How reachable it is, measured three ways

**1. How many patterns must fire.** Over `loadBuiltinRules()`, taking the most generous case for each rule (port bonus applied):

```
rules carrying soft_exclude_patterns that can be dropped : 37
distinct soft patterns that must fire                    : 3   (all 37, no exceptions)
```

Never one, never two. Three.

**2. What stock error pages carry.** Through the shipped resolver:

```
nginx 403 (stock)   patterns fired=1  penalty=0.20  score=0.70  ->  Resolve: "nginx"  0.70
nginx 500 (stock)   patterns fired=1  penalty=0.20  score=0.70  ->  Resolve: "nginx"  0.70
nginx 503 (stock)   patterns fired=1  penalty=0.20  score=0.70  ->  Resolve: "nginx"  0.70
apache 401 (stock)  patterns fired=0  penalty=0.00  score=0.90  ->  Resolve: "Apache" 0.90
```

Every one identifies correctly. The demotion is real; the identification survives it comfortably.

**3. What the validation corpus reaches.** Resolving every banner twice, penalties on and off:

```
probes = 3864    differing = 0    lost entirely = 0    downgraded or changed = 0
```

That zero is **not** evidence the defect is rare, and it must not be read as reassurance. The reachability count is why:

```
corpus banners = 161
[...]

PR #249, opened 2026-08-13
fix(fingerprint): stop protocol hints no rule declares from excluding every candidate

### LINKED ISSUE #240: Five port hints name protocols no rule declares, so 9200/9300/5672/5671/139/135/111 resolve to nothing
## Summary

`detectProtocolFromPort` returns five protocol hints that **no rule in the fingerprint database declares**. When one of those hints reaches the resolver, it is specific enough to switch off fallback mode but matches nothing, so every rule is filtered out and the port resolves to nothing at all. The same banner identifies correctly the moment the hint is empty.

Affected ports: **9200, 9300, 5672, 5671, 139, 135, 111**.

## Mechanism

`pkg/modules/parse/fingerprint_parser.go:539` maps ports to hints:

```go
case 5672, 5671:  return "rabbitmq"
case 9200, 9300:  return "elasticsearch"
case 135:         return "msrpc"
case 139:         return "netbios"
case 111:         return "rpc"
```

`pkg/fingerprint/resolver_rulebased.go:104` gates candidacy on that hint, and only falls back when the hint is generic:

```go
useFallback := in.Protocol == "" || in.Protocol == "tcp" || in.Protocol == "udp"
for _, rule := range r.rules {
    if !useFallback && rule.Protocol != in.Protocol {
        continue
    }
```

The protocol values any rule actually declares:

```
amqp dns ftp http imap kafka ldap memcached mongodb mysql
pop3 postgresql rdp redis smb smtp snmp ssh telnet vnc
```

`rabbitmq`, `elasticsearch`, `msrpc`, `netbios` and `rpc` are not in that set. `rabbitmq` is the clearest case — the rule for it exists and declares `amqp`, so the two sides simply disagree on the name.

## Reproduction

Live against the shipped resolver at `16a5b56`, same banner each time, only the hint differs:

```
port=9200   hint="elasticsearch" declared-by-any-rule=false
   with hint : product=""               err=no matching rule found
   fallback  : product="Elasticsearch"  conf=0.95

port=5672   hint="rabbitmq"     declared-by-any-rule=false
   with hint : product=""               err=no matching rule found
   fallback  : product="RabbitMQ"       conf=0.93

port=139    hint="netbios"      declared-by-any-rule=false
   with hint : product=""               err=no matching rule found
   fallback  : product="Windows SMB"    conf=0.93

port=135    hint="msrpc"        declared-by-any-rule=false
   with hint : product=""               err=no matching rule found
   fallback  : product="Windows SMB"    conf=0.88

port=111    hint="rpc"          declared-by-any-rule=false
   with hint : product=""               err=no matching rule found
   fallback  : product="BIND"           conf=0.85
```

Scratch test in `package fingerprint`:

```go
r := NewRuleBasedResolver(loadBuiltinRules())
in := Input{Banner: "amqp\x00\x00\x09\x01 rabbitmq", Port: 5672}
in.Protocol = "rabbitmq"; fmt.Println(r.Resolve(context.Background(), in))
in.Protocol = "";         fmt.Println(r.Resolve(context.Background(), in))
```

Whether a given scan hits this depends on `protocolHintFor` — the port hint is consulted only when the service name and transport are both generic, which is exactly the case on an unnamed service.

## Two things worth separating

**The five dead hints** are the bug. Either rename `rabbitmq` → `amqp` and drop the four that have no rules, or add the missing protocol values to the rules that serve those ports.

**A guard is missing.** Nothing today asserts that every hint `detectProtocolFromPort` can return is a protocol some rule declares. That check is a few lines over the loaded rule set and it would have caught all five at the moment they were added — it is the more valuable half of the fix, because the hint table and the rule database live in different packages and will drift again.

## Side observation, not the subject

The port 111 fallback row above returns `BIND` at 0.85 for an `rpcbind` banner. `dns.bind`'s pattern is `bind|named|dns`, so the substring `bind` inside `rpcbind` matches. Over-broad single-word patterns are a separate concern from this issue, but that row is a wrong answer rather than a missing one and is worth its own look.

## Notes

Found while auditing the fingerprint tie analysis in #235; separate defect, separate remedy. Measured against `16a5b56`.

cyprobot:
Closes #240.

## The defect, and one instance the issue did not find

A protocol hint naming something no rule declares is worse than no hint at all. It is specific enough to switch the resolver's fallback off, so every rule is filtered out and the port resolves to nothing — while an empty hint puts the resolver in the mode that tries every rule.

#240 found five, all in the port table. There is a sixth, and it arrives through normalization rather than the table:

```
443 / any   https  ->  kept as "https" whenever the response is not HTTP
```

No rule declares `https`. Measured on the shipped resolver:

```
SSH answering on 443       chain hint="https"  with hint: NO MATCH   without: OpenSSH
SMTP answering on 8443     chain hint="https"  with hint: NO MATCH   without: Postfix
real HTTP on 443 (control) chain hint="http"   with hint: nginx      without: nginx
```

The doc comment on `ResolverProtocolHint` already describes this failure in those words — *a hint of "https" matches no rule … and the service then looks unrecognized rather than mis-hinted*. Normalization handled only the half where the response looks like HTTP; the other half returned the value the comment warns about.

## What changed

```
5672/5671  rabbitmq       ->  amqp    the rule exists and declares amqp
9200/9300  elasticsearch  ->  http    both Elasticsearch rules declare http
139        netbios        ->  smb
135        msrpc          ->  (none)  no rule identifies the RPC endpoint mapper
111        rpc            ->  (none)  no rule identifies rpcbind
443/any    https          ->  (none)  when the response is not HTTP
```

Ports 135 and 111 get no hint rather than a renamed one, because no rule identifies those services at all. An empty hint states that; a hint no rule declares states the opposite while excluding everything.

Same banner, before and after, with the port 445 row as the control — a hint that *is* declared, so what differs on 139 is the hint and nothing else:

```
port=9200  hint="elasticsearch"  before: NO MATCH   after: Elasticsearch
port=5672  hint="rabbitmq"       before: NO MATCH   after: RabbitMQ
port=139   hint="netbios"        before: NO MATCH   after: Windows SMB
port=445   hint="smb"            before: Windows SMB  after: Windows SMB   <- control
```

## Why a rule change is in here

Dropping the dead hint on port 111 would have introduced a wrong answer. `dns.bind` matches `bind|named|dns`, so the substring inside `rpcbind` identifies it as **ISC BIND at 0.90** the moment the hint stops excluding every rule. A missing answer becoming a wrong one is a regression for a scanner that feeds product names into CVE correlation, so the narrowing is part of this change rather than a follow-up:

```
match: 'bind|named|dns'  ->  '\bbind|\bnamed|\bdns\b'
```

**And it was overdue independently of port 111.** The same substring match was reporting two other DNS servers as ISC BIND:

```
dnsmasq-2.85                          before: ISC/BIND 0.90    after: not BIND
powerdns authoritative server 4.7.3   before: ISC/BIND 0.90    after: not BIND
rpcbind portmapper v2                 before: ISC/BIND 0.85    after: no match
```

dnsmasq and PowerDNS are DNS servers with their own CVE histories, on far more hosts than rpcbind, and both were being handed to correlation under the wrong product name at higher confidence than the case that prompted this.

The leading boundary does the excluding — `rpcbind` has no word break between `c` and `b`. A trailing boundary after `bind` was in the first version of this branch and cost real coverage: Debian and Ubuntu spell the package `bind9`, version run straight onto it, and that form stopped matching. It is gone. The boundary after `dns` stays, or `dnsmasq` and `powerdns` return.

Swept over the corpus, both rule sets, all ports and hints:

```
probes=1932   verdict changes=0   lost=0   gained=0
control: rpcbind banner  before=ISC/BIND 0.85  after=no match
```

Zero is only informative if the corpus exercises the rule, so that was measured: **four corpus banners match the bind pattern, and all four still match after narrowing**.

But that zero is blind to the thing review caught, and worth saying plainly: all four corpus banners carry the token as a separate word, so **no banner in the corpus distinguishes the two candidate patterns**. The `bind9` regression could not have shown up in it. A correct zero over data that cannot produce the condition — the same shape as the soft-exclude sweep in #239, and the reason the spellings are asserted directly in a test rather than left to the sweep.

## The guard is the more valuable half

Nothing asserted that the hint vocabulary and the rule vocabulary agreed, and they had drifted apart in six places. `fingerprint.DeclaredProtocols()` is added so the check can be written, and the test sweeps **the entire port space** rather than a list of ports somebody thought of — that list is how these got in.

It has to live in `pkg/modules/parse`: that package imports `pkg/fingerprint`, so the same assertion written on the other side is an import cycle.

A second guard covers the normalized hint, since normalization can introduce a value the table never returned — which is exactly what `https` did.

## Two tests asserted the defect

Both are updated, and both are worth noting because they passed the whole time:

- `TestDetectProtocolFromPort` listed the five orphan hints as the expected values.
- `TestFingerprintParserModule_HTTPSResolverProtocolNotNormalizedForNonHTTPResponse` asserted the resolver is handed `https` for a non-HTTP response. **Its own fixture is an SMTP greeting on port 443** — the case being lost was sitting inside the test that locked the behaviour in.

## Verification

| check | result |
|---|---|
| `go test ./...` | passes |
| `golangci-lint run ./...` (v2.12.2, CI-pinned) | 0 issues |
| `go vet` on both touched packages | clean |
| mutation: restore any of the six hints | guard fails |
| mutation: restore `bind|named|dns` | guard fails |

Every mutation was checked for compiling before its result was read — a mutation that does not build reads exactly like one nothing catches.

## Not included

The `.golangci.yaml` and probe-catalog vocabularies are untouched. `data/probes.yaml` declares `msrpc`, `netbios` and `rpc` as probe protocols; those are probe identifiers rather than resolver hints and no longer reach the resolver after this change, but the same drift could recur there and it is not covered by this guard.

### DIFF
--- a/pkg/fingerprint/data/fingerprint_db.yaml
+++ b/pkg/fingerprint/data/fingerprint_db.yaml
@@ -922,6 +922,11 @@ rules:
     product: 'BIND'
     vendor: 'ISC'
     cpe: 'cpe:2.3:a:isc:bind:*:*:*:*:*:*:*:*'
+    # The leading boundaries keep rpcbind, dnsmasq and powerdns from being
+    # reported as BIND; there is deliberately none after 'bind', or the
+    # bind9 spelling stops matching. Guarded by
+    # TestDNSIdentificationSurvivesTheNarrowing in pkg/modules/parse, which
+    # cannot live here: that package imports this one.
     match: '\bbind|\bnamed|\bdns\b'
     version_extraction: "(?:bind|named)[\\s/_-]+([\\d\\.]+)"
 
--- a/pkg/fingerprint/data/fingerprint_db.yaml
+++ b/pkg/fingerprint/data/fingerprint_db.yaml
@@ -922,7 +922,7 @@ rules:
     product: 'BIND'
     vendor: 'ISC'
     cpe: 'cpe:2.3:a:isc:bind:*:*:*:*:*:*:*:*'
-    match: 'bind|named|dns'
+    match: '\bbind\b|\bnamed\b|\bdns\b'
     version_extraction: "(?:bind|named)[\\s/_-]+([\\d\\.]+)"
 
     exclude_patterns:
--- a/pkg/fingerprint/data/fingerprint_db.yaml
+++ b/pkg/fingerprint/data/fingerprint_db.yaml
@@ -922,7 +922,7 @@ rules:
     product: 'BIND'
     vendor: 'ISC'
     cpe: 'cpe:2.3:a:isc:bind:*:*:*:*:*:*:*:*'
-    match: '\bbind\b|\bnamed\b|\bdns\b'
+    match: '\bbind|\bnamed|\bdns\b'
     version_extraction: "(?:bind|named)[\\s/_-]+([\\d\\.]+)"
 
     exclude_patterns:
--- a/pkg/fingerprint/loader.go
+++ b/pkg/fingerprint/loader.go
@@ -8,6 +8,7 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"sort"
 )
 
 //go:embed data/fingerprint_db.yaml
@@ -24,6 +25,30 @@ func loadBuiltinRules() []StaticRule {
 		return nil
 	}
 	return prepareRules(rules)
+}
+
+// DeclaredProtocols returns, sorted, every protocol value the shipped rules
+// declare.
+//
+// It exists so that anything producing a protocol hint can check its
+// vocabulary against the rules' rather than the two drifting apart in silence.
+// The failure is not a missed match but a total one: a hint no rule declares
+// is still specific enough to switch fallback off, so it excludes every
+// candidate and the service resolves to nothing at all.
+func DeclaredProtocols() []string {
+	seen := make(map[string]struct{})
+	for _, rule := range loadBuiltinRules() {
+		if rule.Protocol != "" {
+			seen[rule.Protocol] = struct{}{}
+		}
+	}
+
+	protocols := make([]string, 0, len(seen))
+	for protocol := range seen {
+		protocols = append(protocols, protocol)
+	}
+	sort.Strings(protocols)
+	return protocols
 }
 
 // loadExternalCatalog attempts to load fingerprint rules from workspace cache.
--- a/pkg/modules/parse/fingerprint_parser.go
+++ b/pkg/modules/parse/fingerprint_parser.go
@@ -478,10 +478,16 @@ func normalizeResolverProtocol(protocolHint, response, probeID string) string {
 	if protocolHint != "https" {
 		return protocolHint
 	}
-	if !looksLikeHTTPResponse(response, probeID) {
-		return protocolHint
+	if looksLikeHTTPResponse(response, probeID) {
+		return "http"
 	}
-	return "http"
+	// No rule declares https, and a hint no rule declares is still specific
+	// enough to switch fallback off -- so returning it here excluded every
+	// candidate. What is actually known at this point is that the service
+	// answered over TLS with something that is not HTTP, which says nothing
+	// about the protocol; an empty hint says exactly that and lets every rule
+	// compete. SSH on 443 and SMTP on 8443 both resolve to nothing otherwise.
+	return ""
 }
 
 func looksLikeHTTPResponse(response, probeID string) bool {
@@ -565,23 +571,25 @@ func detectProtocolFromPort(port int) string {
 	case 389, 636, 3268, 3269:
 		return "ldap"
 	case 5672, 5671:
-		return "rabbitmq"
+		// The rule for RabbitMQ declares amqp; naming the product here instead
+		// of the protocol filtered every rule out and resolved the port to
+		// nothing.
+		return "amqp"
 	case 9092, 9093:
 		return "kafka"
 	case 9200, 9300:
-		return "elasticsearch"
+		// Elasticsearch answers over HTTP and both of its rules declare http.
+		return "http"
 	case 161, 162:
 		return "snmp"
 	// File Sharing / Windows Services
-	case 135:
-		return "msrpc"
-	case 139:
-		return "netbios"
-	case 445:
+	// 135 is the RPC endpoint mapper and no rule identifies it. An empty hint
+	// says that honestly and lets every rule compete; a hint no rule declares
+	// says the opposite while excluding all of them.
+	case 139, 445:
 		return "smb"
 	// Infrastructure Services
-	case 111:
-		return "rpc"
+	// 111 is rpcbind, which no rule identifies either. See 135 above.
 	}
 	return ""
 }
--- a/pkg/modules/parse/fingerprint_parser_test.go
+++ b/pkg/modules/parse/fingerprint_parser_test.go
@@ -358,18 +358,26 @@ func TestDetectProtocolFromPort(t *testing.T) {
 		{"LDAPS secure port", 636, "ldap"},
 		{"LDAP global catalog", 3268, "ldap"},
 		{"LDAP global catalog SSL", 3269, "ldap"},
-		{"RabbitMQ standard port", 5672, "rabbitmq"},
-		{"RabbitMQ secure port", 5671, "rabbitmq"},
+		// These four named the product or the service rather than the
+		// protocol the rules declare, and this test asserted that as correct.
+		// A hint no rule declares excludes every candidate, so each of them
+		// meant the port resolved to nothing at all.
+		{"RabbitMQ standard port hints its rule's protocol", 5672, "amqp"},
+		{"RabbitMQ secure port hints its rule's protocol", 5671, "amqp"},
 		{"Kafka standard port", 9092, "kafka"},
 		{"Kafka secure port", 9093, "kafka"},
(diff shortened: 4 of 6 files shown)


cyprobot (commented):
The sixth hint is the right find and it is the one I missed. `https` is not in the rules' vocabulary — I measured that set earlier today and it is `amqp dns ftp http imap kafka ldap memcached mongodb mysql pop3 postgresql rdp redis smb smtp snmp ssh telnet vnc`, twenty values with no `https` among them. So every TLS service that is not HTTP was resolving to nothing, which is a larger population than the five ports I filed.

Bundling the `dns.bind` narrowing is also right, and for the reason you give rather than as a convenience: dropping 111's hint is what lets `rpcbind` reach the rules, so the regression is yours to prevent in the same change. Turning "no answer" into "wrong answer" would have been the worse trade, and a product name feeding CVE correlation makes it worse still.

## One finding: the trailing `\b` on `bind` costs coverage and buys nothing

```
                                                  old    PR     proposed
bind 9.11.4-p2-redhat                             true   true   true
isc bind 9.18.24                                  true   true   true
named 9.16.1                                      true   true   true
named-9.16.1                                      true   true   true
dns server ready                                  true   true   true
bind9                                             true   FALSE  true     <<<
bind9.16.1                                        true   FALSE  true     <<<
rpcbind 2.5                                       true   false  false
100000 rpcbind portmapper program version         true   false  false
dnsmasq-2.85                                      true   false  false
powerdns authoritative server 4.7.3               true   false  false
unbound 1.13.1                                    false  false  false
```

`\bbind\b` requires a non-word character after `bind`, so `bind9` — the Debian and Ubuntu package spelling, and a form that turns up in real banners — stops matching. `bind9.16.1` goes with it.

The exclusion you want comes entirely from the **leading** boundary: `rpcbind` has no word boundary between `c` and `b`, so `\bbind` alone already refuses it. The trailing one only removes version-suffixed spellings.

```yaml
match: '\bbind|\bnamed|\bdns\b'
```

Twelve cases, two wrong with the PR's form, zero wrong with this one. Keep the trailing `\b` on `dns` — without it `dnsmasq` and `powerdns` come straight back.

**Your sweep cannot see this, and the reason is the one from #239.** The corpus holds four banners that wake this rule — `BIND 9.16.23`, `BIND 9.16.1-Ubuntu`, `named version 9.18.4`, `DNS service running` — and every one of them has the token as a standalone word. So all four survive the narrowing, your reachability count is honest, and the zero is real. It is also blind to the one shape that regresses, because the corpus contains no `bind9`-style banner. Third time today that a correct zero was measured over a corpus that cannot produce the condition.

## The narrowing fixes more than the PR claims

`dnsmasq` and `powerdns` were both being reported as **ISC BIND 0.85**. Neither is BIND. They are separate DNS servers with their own CVE histories, and under the old pattern both were feeding the wrong product name into correlation — the same defect as `rpcbind`, on hosts far more common than an RPC portmapper.

That is a correctness gain nobody counted. Worth putting in the body: it turns "I had to bundle a rule change" into "the rule change was overdue and this PR is where it surfaced".

## Agreed without reservation

- **`DeclaredProtocols()` and the guard in `parse`.** The import direction settles it — `parse` already takes `fingerprint`, so the assertion cannot live on the other side. Sweeping the whole port space rather than a list of remembered ports is the part that matters; a hand-maintained list is how the five got in.
- **135 and 111 get no name rather than a new one.** No rule identifies either service, and an empty hint says that honestly while a made-up one says the opposite and excludes everything.
- **The two tests that had locked the defect in.** The second is the one worth remembering: the fixture asserting `https` was an SMTP greeting on 443 — the lost case was sitting inside the test that preserved the behaviour. That is not a bad test, it is a test whose fixture was the bug.
- **Your mutation harness reverting your own fix via `git checkout`**, caught by `git diff --stat`. Same family as everything else today, and you caught it before publishing.

## Out of scope, agreed

`data/probes.yaml` declaring `msrpc`/`netbios`/`rpc` as probe identifiers is a different namespace and does not reach the resolver after this change. Worth its own issue only if the two vocabularies can drift into each other again — which this guard does not cover, as you say.

## Verdict

One line to change: drop the trailing `\b` after `bind`. Everything else stands, and the `https` case makes this a materially bigger fix than the issue I filed.

Merge is not mine.

cyprobot (commented):
Verified at `144d61f` rather than taking it: pattern is `\bbind|\bnamed|\bdns\b`, `TestDNSIdentificationSurvivesTheNarrowing` passes, and it fails in both directions — restoring the trailing boundary loses `bind9` and fails it; removing the narrowing entirely makes `dnsmasq` resolve as BIND and fails it. Both mutations caught.

The `bindfs` trade is right and worth the sentence you gave it: a FUSE filesystem does not appear in a network banner, `bind9` is the standard Debian spelling, and the asymmetry decides it.

**One small thing, not blocking.** The test lives in `pkg/modules/parse` while its subject is a single line of `pkg/fingerprint/data/fingerprint_db.yaml`. That placement is forced — same import direction that put `DeclaredProtocols()` where it is — but it means someone editing that rule will not see the guard from where they are editing. A pointer comment beside the pattern ("narrowing guarded by TestDNSIdentificationSurvivesTheNarrowing in pkg/modules/parse") costs nothing and stops the next person from re-broadening it and finding out from an unrelated package's test.

The rewritten body is better than the original framing. `dnsmasq` and PowerDNS resolving as ISC BIND at **0.90** — higher confidence than the `rpcbind` case that motivated the change, on far more common hosts — is the strongest fact in this PR and it now reads that way.

Nothing further from me.

PR #255, opened 2026-08-14
fix(fingerprint): unescape imap.generic — its plain-greeting branch has never matched anything

### LINKED ISSUE #241: imap.generic is double-escaped: the branch that catches a plain IMAP greeting is dead
## Summary

`imap.generic`'s pattern is double-escaped. Its first alternative matches literal backslashes instead of whitespace, so the branch written to catch a plain IMAP greeting never fires. The rule only ever matches through its second alternative, which requires the banner to contain `capability`.

## The pattern

`pkg/fingerprint/data/fingerprint_db.yaml:649`

```yaml
- id: 'imap.generic'
  protocol: 'imap'
  description: 'Generic IMAP server (Dovecot, Courier, Cyrus, Gmail, etc.)'
  match: '\\*\\s+ok\\s+.*imap|capability.*imap4'
```

Single-quoted YAML does not process escapes, so the regex the engine compiles contains `\\*` and `\\s` — an escaped backslash followed by a quantifier, not an escaped star and a whitespace class.

## Reproduction

```go
re := regexp.MustCompile(`(?i)` + `\\*\\s+ok\\s+.*imap|capability.*imap4`)
```

```
"* ok imap4rev1 server ready"                 -> false
"* ok dovecot ready"                          -> false
"* ok [capability imap4rev1] server ready"    -> true      <- second alternative only
`\*\s\sok\s\simap`                            -> true      <- what it actually matches
```

With a single backslash, which is what the description implies was intended:

```
"* ok imap4rev1 server ready"                 -> true
```

## Effect

An IMAP server whose greeting is the ordinary `* OK IMAP4rev1 ...` form is not matched by the generic rule. It is still caught when the greeting names a specific product (`imap.dovecot`, `imap.smartermail`) or when the banner happens to carry `capability` — which it does when the collector issued a `CAPABILITY` command, and does not when it only read the greeting.

So the gap is narrow but real, and it is invisible: the rule exists, its description promises Dovecot/Courier/Cyrus/Gmail coverage, and nothing reports that half of it is unreachable.

## Fix

Single-escape the first alternative:

```yaml
match: '\*\s+ok\s+.*imap|capability.*imap4'
```

Worth adding, in the same change, a database-wide assertion that no `match` pattern contains `\\\\` — this is the kind of defect that is invisible on review and trivial to detect mechanically. A scan of the current database for the same shape would say whether `imap.generic` is the only one.

## Notes

Found while constructing witness banners during the fingerprint tie audit in #235; unrelated to that issue. Measured against `16a5b56`.

cyprobot:
Closes #241.

## The defect

`pkg/fingerprint/data/fingerprint_db.yaml:649` holds the pattern in a **single-quoted** YAML scalar, which does not process escapes, but it was escaped as though it were double-quoted. Every escape in the first alternative arrived at the regex engine doubled — an escaped backslash followed by a quantifier, where an escaped star and a whitespace class were meant.

Both doubled escapes are in the **first** alternative, so that branch could only match a banner containing a literal backslash. No IMAP greeting does.

## Why nobody noticed

The rule kept matching through its second alternative, which requires the word `capability`. So servers that advertise CAPABILITY in their greeting were identified correctly, and only the ones the first branch was written for were missed — which is exactly the set the rule's description names: *"Dovecot, Courier, Cyrus, Gmail"*. Those commonly answer with a plain `* OK` and wait to be asked.

A rule that half works is harder to see than one that does not work at all.

Measured against the shipped pattern, before and after:

```
"* OK IMAP4rev1 Server ready"          before: no match    after: match
"* OK [CAPABILITY IMAP4rev1] ready"    before: match       after: match
```

## Scope check

The same mistake could be anywhere in the file, so I looked rather than assumed:

```
50 rules total
42 single-quoted patterns  -> exactly 1 contains a doubled backslash (this one)
 8 double-quoted patterns  -> doubling is correct there and all 8 are correct
```

So this is a single slip, not a convention problem, and no other rule needs the same change.

## Tests

`pkg/fingerprint/imap_generic_test.go`, asserting in both directions:

- the revived branch matches a plain greeting, through the resolver and against the raw pattern
- the branch that already worked still does
- **a server that names itself still wins**: the Dovecot greeting resolves to `Dovecot`, not to the generic rule. Widening a generic pattern risks trading a missed identification for a wrong one, which is worse, so this is asserted rather than assumed
- the wider pattern does not claim `+OK POP3` or an SMTP `220` greeting

**Mutation-checked.** Restoring the doubled escapes fails three of them:

```
--- FAIL: TestIMAPGeneric_PlainGreetingIsIdentified
--- FAIL: TestIMAPGeneric_PatternIsNotDoubleEscaped
--- FAIL: TestIMAPGeneric_ResolvesUnderBothHintAndFallback
```

One of these tests corrected me rather than the code: it was first written expecting the Dovecot banner to resolve to `IMAP`, and the resolver was right to prefer the specific rule. The assertion now records that precedence instead of contradicting it.

## Verification

- `go test ./...` — clean
- `golangci-lint run ./pkg/fingerprint/...` — 0 issues
- `vulntor fingerprint validate pkg/fingerprint/data/fingerprint_db.yaml` — exit 0

Note that `Fingerprint Validation` will run on this PR and compare against the `main` baseline, since the change touches `pkg/fingerprint/data/`. That comparison is the useful signal here — the point of the change is that a set of banners resolves differently than it did.

### DIFF
--- a/pkg/fingerprint/data/fingerprint_db.yaml
+++ b/pkg/fingerprint/data/fingerprint_db.yaml
@@ -646,7 +646,7 @@ rules:
     product: 'IMAP'
     vendor: 'Various'
     cpe: 'cpe:2.3:a:*:imap:*:*:*:*:*:*:*:*'
-    match: '\\*\\s+ok\\s+.*imap|capability.*imap4'
+    match: '\*\s+ok\s+.*imap|capability.*imap4'
     version_extraction: ''
 
     soft_exclude_patterns:
--- a/pkg/fingerprint/imap_generic_test.go
+++ b/pkg/fingerprint/imap_generic_test.go
@@ -1,0 +1,120 @@
+package fingerprint
+
+import (
+	"context"
+	"regexp"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// Greetings as the servers actually send them. An IMAP session opens with an
+// untagged `*` line, and what follows it is the server's choice: RFC 3501
+// suggests advertising CAPABILITY there, and some do, but Dovecot and Courier
+// commonly answer with a plain OK and wait to be asked.
+//
+// That split is the whole point of imap.generic having two alternatives. The
+// second one requires the word "capability"; the first exists to catch the
+// servers that do not send it.
+const (
+	imapPlainGreetingBanner      = "* OK IMAP4rev1 Server ready\r\n"
+	imapDovecotGreetingBanner    = "* OK Dovecot (Ubuntu) ready. IMAP4rev1\r\n"
+	imapCapabilityGreetingBanner = "* OK [CAPABILITY IMAP4rev1 STARTTLS] Server ready\r\n"
+)
+
+// The regression this file exists for.
+//
+// The pattern was written in a single-quoted YAML scalar, which does not
+// process escapes, but escaped as though it were double-quoted. The engine
+// therefore compiled `\\*` — an escaped backslash followed by a quantifier —
+// where `\*` was meant, and likewise `\\s` for `\s`. Both live in the FIRST
+// alternative, so that entire branch could only match a banner containing a
+// literal backslash, and no IMAP greeting does.
+//
+// The rule still matched through its second alternative, which is why nothing
+// looked broken: servers that advertise CAPABILITY in the greeting were
+// identified, and only the ones the first branch was written for were missed.
+// A rule that half works is harder to notice than one that does not work.
+func TestIMAPGeneric_PlainGreetingIsIdentified(t *testing.T) {
+	result := requireResolved(t, imapPlainGreetingBanner, "imap", 143)
+
+	require.Equal(t, "IMAP", result.Product,
+		"the first alternative of imap.generic exists for greetings that do "+
+			"not advertise CAPABILITY; if this fails it is matching nothing")
+}
+
+// A banner that names its server belongs to that server's rule, not to the
+// generic one. Asserted because the fix widens what imap.generic can match, and
+// a generic rule that starts winning against specific ones would trade a missed
+// identification for a wrong one — the worse of the two.
+//
+// This is also the case that corrected the test rather than the code: it was
+// first written expecting "IMAP" here, and the resolver was right.
+func TestIMAPGeneric_DoesNotOutrankAServerThatNamesItself(t *testing.T) {
+	result := requireResolved(t, imapDovecotGreetingBanner, "imap", 143)
+
+	require.Equal(t, "Dovecot", result.Product,
+		"the dovecot rule is more specific and must keep winning")
+}
+
+// The branch that was already working must keep working: the fix removes
+// escaping from the first alternative and must not disturb the second.
+func TestIMAPGeneric_CapabilityGreetingStillIdentified(t *testing.T) {
+	result := requireResolved(t, imapCapabilityGreetingBanner, "imap", 993)
+
+	require.Equal(t, "IMAP", result.Product)
+}
+
+// Asserted against the shipped pattern rather than through the resolver,
+// because this is the defect itself: a pattern that cannot match what it
+// describes. Going through Resolve would report the same failure, but not say
+// which of the two alternatives died.
+func TestIMAPGeneric_PatternIsNotDoubleEscaped(t *testing.T) {
+	rule := ibmRule(t, "imap.generic")
+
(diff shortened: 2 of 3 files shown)


cyprobot (commented):
Took the three things you flagged in reverse order, because the third is the one you could not close yourself.

## 3. The sweep is complete, and I checked the sweeper before believing it

You looked at `match:` only. I swept every regex-bearing field on the rule struct — `match`, `version_extraction`, `exclude_patterns`, `soft_exclude_patterns` — 338 values in total, 47 single-quoted and 291 double-quoted.

```
main (fix absent)   fields scanned=338   findings=1
    -> line 649  imap.generic [match]  \\*\\s+ok\\s+.*imap|capability.*imap4
PR #255             fields scanned=338   findings=0
```

Same 338 both times, and the detector finds the known defect on `main` — so the zero on your branch measures something rather than reporting itself. `imap.generic` was the only instance in the database, in any field.

Worth knowing why the double-quoted 291 are safe by construction: a double-quoted YAML scalar containing a single `\s` is an invalid escape, so it fails at parse time rather than shipping wrong. The defect shape only exists in single-quoted scalars, which is 47 of the 338.

## 1. The widening does not steal anything, measured wider than your list

```
Dovecot IMAP greeting      -> Dovecot          0.95     specific rule still wins
plain IMAP greeting        -> IMAP             0.90     the branch that was dead
Courier IMAP               -> IMAP             0.90
Cyrus IMAP                 -> IMAP             0.90
POP3 greeting              -> (no match)
SMTP greeting              -> Postfix          0.93
"* OK news server ready"   -> (no match)              <- `* OK` without imap
HTTP page mentioning imap  -> nginx            0.90   <- imap without `* OK`
FTP                        -> ProFTPD          0.93
SSH                        -> OpenSSH          0.95
```

The two shapes that could plausibly collide are the last ones I added: a banner with `* OK` and no `imap`, and a banner with `imap` and no `* OK`. Neither is taken, because the alternative requires both. Courier and Cyrus — two of the three servers the rule's own description names — now resolve where they did not.

Your test correcting you rather than the code is the right outcome and the right instinct to be nervous about. The risk you named is real: widening a generic pattern trades a missed identification for a wrong one, and a more specific rule losing its precedence is how that happens. It holds here at 0.95 against 0.90, and the assertion you changed now pins that ordering, which is what it should assert.

## 2. On the dataset case — I would not split it

You ask whether adding the missing validation case widened the PR's scope. It did, and it should have.

A fix whose corpus cannot exercise it is the exact pattern this repository has produced four times today: the tie guard that finds no ties, the soft-penalty sweep over banners with no error pages, the DNS narrowing invisible to a corpus with no `bind9` spelling. Split this one and you get a fix whose measured effect is zero — indistinguishable from a fix that does nothing — plus a dataset PR whose reason lives in another thread.

The F1 sequence is what makes it load-bearing rather than decorative: `0.9099` broken, `0.9138` before, `0.9145` after. It moves in both directions, so it is a case that measures rather than a case that passes.

Correcting the "the baseline comparison will be the real signal" claim once you found it was not is the more important half of that anyway.

## The metrics script finding is real — reproduced

```
percent_change 100.00 100.00  ->  "0.00"     guard tests == "0"   ->  NO MATCH
percent_change 0      0       ->  "0"        guard tests == "0"   ->  match
```

The literal `"0"` only ever comes from the divide-by-zero path. Every real metric goes through `printf("%.2f")` and returns `"0.00"`, which the string comparison misses — so an unchanged metric falls into the changed branches and, for a higher-is-worse metric, is labelled `❌ Regressed` with `regressed=1` set.

It is invisible today because the hard-fail threshold is 5%, and it becomes a build break for every PR the moment anyone tightens that to zero — which is the natural thing to do once someone trusts the table. Right call to keep it out of this PR; worth its own issue with the two lines above in it, because the fix is a numeric comparison rather than a string one and that is a one-line change somebody will otherwise re-derive.

## One of mine, since I have been holding everyone to it

My first attempt at the widening table silently did nothing: I chained `grep ... && cat > test.go <<EOF`, the grep matched nothing, the `&&` short-circuited, the file was never written, and `go test` said `no tests to run` — which I read as a result for a moment before noticing the file did not exist. Same family as the harness that reverted its own fix. The tell was the same one too: check that the thing you think you wrote is actually there.

## Verdict

Fix is correct, the sweep behind it is now exhaustive, and the widening is measured rather than argued. Nothing further from me.

Merge is not mine — it goes to hq.

Issue #248, opened 2026-08-13
A required status check is named after the Go version, so every Go bump blocks itself


cyprobot:
## Summary

One required status check is named after a matrix value that includes the Go version, so **any Go version bump blocks itself**. The PR that changes the version renames the job, the required context never appears, and the PR sits `BLOCKED` with every check green.

This is not hypothetical — it happened today on #246 and cost a branch-protection edit to clear.

## What happened

`#246` bumped the build matrix from `1.25` to `1.26.6`. All nine checks passed. The PR was still unmergeable:

```
mergeStateStatus = BLOCKED
checks            = SUCCESS × 9

required context : Build Binary (ubuntu-latest, 1.25)     <- never appears again
actually ran     : Build Binary (ubuntu-latest, 1.26.6)
```

## Why

`test-build` in `.github/workflows/test.yaml` is a matrix job:

```yaml
  test-build:
    name: Build Binary
    strategy:
      matrix:
        os: [ubuntu-latest, macos-latest]
        go-version: ['1.26.6']
```

GitHub appends the matrix values to the job name, so the status context is `Build Binary (ubuntu-latest, <go-version>)` — the Go version is part of the check's identity. Branch protection matches required checks by name, so the moment the version changes, the required one becomes unreachable.

The failure is silent in the way that matters: nothing says "this required check can no longer run". It reads as an ordinary pending gate, and the only ways past it are editing branch protection or merging with admin — the second of which is how a version bump ends up bypassing the gate it was supposed to satisfy.

Every future Go bump does this again. So does any change to the `os` list.

## Immediate state

I swapped the required context to `Build Binary (ubuntu-latest, 1.26.6)` so #246 could merge on its own checks rather than by admin override. That unblocked it and **moved the problem to the next bump** rather than fixing it.

## Fix

Take the version out of the check's identity. Either:

- **Drop `go-version` from the matrix** and read it from `env.GO_VERSION` (or `go.mod`) inside the job. There is only one value in that list today, so the matrix dimension buys nothing and costs this. The context becomes `Build Binary (ubuntu-latest)`.
- **Or give the job an explicit stable `name`** that does not interpolate the matrix, and require that.

Then update the required context once more, to the stable name, and it stops recurring.

## Worth noting while here

[...]

PR #260, opened 2026-08-17
fix: rename cmd/vulntor to cmd/cyprob and fix the version-injection ldflags they broke

cyprobot:
## Problem

The module moved to `github.com/cyprob/cyprob`, but the CLI's own package path (`cmd/vulntor/`), binary name, version-injection ldflags, and nearly every user-facing string the CLI prints still said `vulntor`. #258 asked for the code-side rename: `cmd/vulntor/` → `cmd/cyprob/`, `BIN_NAME`, the gci prefix, and the ldflags — the last one flagged as a real defect, not cosmetics.

Measured: `-X` in the Makefile targets `github.com/vulntor/vulntor/pkg/version`, a package path that no longer exists under this module. Go does not error on an unmatched `-X` flag — it silently injects nothing. Every binary this Makefile has produced since the module rename reports a fixed `v0.5.3+unknown` instead of its real version, commit, and build date.

## User / Ops Impact

Anyone building a release binary from this Makefile gets one that cannot report what it is — `vulntor version` always says `v0.5.3+unknown` regardless of what was actually built. `test.yaml`'s `Build Binary` job already runs `./dist/.../vulntor version` on every push, and passes every time, because it checks that the command exits zero, not what it prints — so this has been silently wrong on every CI run since the rename and nothing caught it. This PR does not add that check; it fixes what the existing check would have caught if it looked at its own output.

## Fix Summary

`cmd/vulntor/` → `cmd/cyprob/` (all internal import paths + `.golangci.yaml`'s path exclusions), `Makefile`'s `BIN_NAME`/gci prefix/`-X` ldflags, and `cliExecutable` in `root.go`.

That last one is why this touches far more than four lines: renaming `BIN_NAME` alone would ship a binary named `cyprob` whose own `--help` claims its name is `vulntor` — a broken half-rename. Fixing `cliExecutable` pulls in everything downstream of the CLI's self-reported name: every subcommand's `Example`/`Short`/`Long` help text, the error-hint strings printed on failure, a startup log line, and doc comments describing "the Vulntor application" / "the Vulntor server". All of it is listed in the commit message rather than repeated here.

**The riskiest part of doing this by hand:** a full-repo sweep after the first pass looked complete turned up a second, independent copy of every error hint in `cmd/cyprob/internal/format/summary.go` — the one the CLI's `PrintTotalFailureSummary` actually calls — that the first pass missed entirely, plus several `Short`/`Long` strings a batch `sed` reported as "fixed" that never actually changed, because BSD `sed`'s `\b` silently doesn't behave like GNU `sed`'s. Caught only by re-grepping after, not by trusting the script's own success output — recorded in the commit message as its own lesson, since it's the second time today an assumption about which `sed`/`grep` dialect is running produced a false "done."

## Behavior Change

`cyprob version`, `cyprob --help`, and every subcommand's `--help` now consistently say `cyprob`. `DefaultWorkspaceRoot()`'s default moves from `~/.local/share/vulntor` to `~/.local/share/cyprob` (and the macOS/Windows equivalents), **with a one-release fallback**: if the new path doesn't exist yet but the old one does, the old one is used — so upgrading does not make an existing install's scan history and config look like it vanished. `pkg/paths` (`ConfigDir`/`DataDir`/`CacheDir`) gets the same rename without a fallback, since `ConfigDir`/`DataDir` have no callers anywhere in the codebase today and `CacheDir` holds only re-downloadable plugin cache.

No plugin manifest schema, no environment variable name, and no wire-visible network identifier changes — see below.

## Evidence

```
go build ./...            clean
go test ./...             clean
golangci-lint run ./...   0 issues
gofmt -l                  clean, except pre-existing drift already on origin/main (diffed to confirm)
```

Verified against the **built binary**, not just the source, after each pass:

```
$ ./cyprob version                          # old ldflags path
vulntor version: v0.5.3+unknown             # wrong, every time, regardless of what was built

$ ./cyprob version                          # this PR's ldflags path
cyprob version: vunknown+727b39c            # injects the real commit

$ ./cyprob scan
💡 Suggestions:
  → Provide a target:           cyprob scan 192.168.1.0/24

$ ./cyprob storage --help
Manage Cyprob storage operations including garbage collection.
```

The workspace-root fallback is exercised directly via `t.Setenv("HOME", ...)` against a temp directory rather than relying on whatever happens to exist on the machine running the test: neither path exists (fresh install → new path), only the old path exists (→ fallback), both exist (→ new path, already migrated), and a file rather than a directory at the old path (→ not mistaken for a workspace, new path). **Mutation-checked**: removing the fallback fails that suite.

The `error-handling` CI gate's `path: cmd/vulntor/commands/plugin` is `required: true` and enforced — and `check-error-handling.sh` treats a missing path as `✅ No violations found (path does not exist)`, exit 0. Left stale, that gate would have gone green forever by checking nothing. Confirmed by reading the script, not assumed.

## Deliberately not touched, and why

- **`pkg/plugin/types.go`'s `vulntor_min_version` YAML/JSON manifest field, `MinVulntorVersion`, `IsCompatibleWithVulntor`.** Public plugin-manifest schema — renaming it breaks every plugin repository that writes this field (`cyprob-plugin-source`, `cyprob-nuclei-plugin-source`, `cyprob-ee-plugin-source`, third-party content). Needs its own decision. The prose next to it ("plugin requires Vulntor >= ...") is left saying Vulntor too — renaming only the sentence around an untouched field name would be more confusing, not less.
- **`Author: "Vulntor Team"` / `"vulntor-security"` / `"vulntor-core"`** — plugin and native-module author metadata, shipped and in test fixtures. Content attribution, not code plumbing.
- **`// Copyright 2025 Vulntor Authors`** — present in 61 files. Legal attribution, not mine to change unilaterally.
- **The User-Agent header `banner_grab.go` sends while probing targets** (`vulntor-probe/1.0`) — the one string in this change that actually leaves the process and is visible to whatever the scanner points at. Some deployment may already key a firewall rule, IDS signature, or dashboard off that exact string. Worth a deliberate decision, not a silent rename either way.
- **`VULNTOR_*` as the environment-variable prefix the whole application reads its config from** (`pkg/config/source.go`'s `EnvSource`, `VULNTOR_WORKSPACE` and siblings on `storage.Config`). Found while checking whether the workspace-root env override needed updating too — it doesn't, only the computed default changed. This prefix is wired into **cyprob-ee's production deployment today**: docker-compose files (dev, HA, prod, remote, the ephemeral preview environments), `.env.example` files, and EE's own config source all set `VULNTOR_*` variables. This is the single largest, most consequential rename surface anywhere in this issue, and nobody asked for it. It needs cross-repo coordination and its own decision — flagging prominently rather than including a line for it in this diff.
- **`.github/workflows/sync-docs.yml`** pushes to `https://github.com/vulntor/docs.git`. Confirmed via the API: `vulntor/docs` no longer exists, `cyprob/docs` does. Confirmed via run history: this workflow has failed on every trigger since **2025-11-14**, over nine months, unnoticed. A real, separate, already-broken pipeline — not fixed here because "point it at `cyprob/docs`" isn't obviously right without knowing why nobody already did that.

## Risk / Rollback

Moderate diff size (86 files), low behavioral risk: almost all of it is help text, error-message strings, and doc comments — text a user reads, not logic a program branches on. The two places with real behavioral surface are `DefaultWorkspaceRoot()` and `pkg/paths`, both covered above with a fallback and dedicated tests.

Reverting the commit restores every renamed string and path exactly; there is no data migration or schema change to un-migrate. The workspace-root fallback means even a partial rollback (reverting the code but not un-migrating any data) is safe — an old-named workspace directory is still found and used.

## Linked Issue

Refs #258


cyprobot (commented):
Every claim in this PR is checked against something running or something on GitHub, not against the prose describing it. All of it held.

## The defect this PR exists to fix — reproduced directly on the built binary

Bypassed `make binary` (fails on an unrelated pnpm build-approval prompt for `@swc/core`/`esbuild` — pre-existing, not this PR's problem) and built the Go binary straight with the Makefile's own ldflags:

```
$ cyprob version
cyprob version: vtest+0020493
Commit: 00204931d50e593ec6fc1a5d3c4e024689cedd70
```

Real commit injected. Confirmed the PR's own repro of the *old* bug is real too: the old `-X github.com/vulntor/vulntor/pkg/version.*` path doesn't exist under this module, Go silently no-ops an unmatched `-X`, and that's a defect independent of this rename — this PR is what makes it visible and fixes it in the same motion.

`--help`, `scan` (no targets), `storage --help` — all render as `Cyprob`/`cyprob` consistently, matching the Evidence section exactly.

## The one real behavioral surface — read, then broken on purpose

`DefaultWorkspaceRoot()`'s new/old-path fallback: read it end to end, ran `TestDefaultWorkspaceRoot_MigrationFallback`'s four subtests (clean), then removed the fallback branch myself and reran:

```
--- FAIL: TestDefaultWorkspaceRoot_MigrationFallback/only_the_old_path_exists:_falls_back_to_it
```

Catches exactly the case that matters — an existing install losing its workspace on upgrade. `dirExists` correctly requires `IsDir()`, so a stray file at the old path can't be mistaken for a workspace, and `VULNTOR_WORKSPACE`'s env-tag on `Config.WorkspaceRoot` is untouched in the diff, confirmed by reading it directly.

## Every "measured" claim in the body, re-measured

- **ldflags / version injection** — above.
- **Error-handling gate fails open on a missing path** — read `.github/scripts/check-error-handling.sh` directly: `[ ! -d "$PATH_TO_SCAN" ]` → `"✅ No violations found (path does not exist)"`, `exit 0`. Confirmed `cmd/cyprob/commands/plugin` exists and `validate.yaml`'s `path:` now points there — a required gate that would otherwise have gone green by checking nothing.
- **`sync-docs.yml` → `vulntor/docs` broken, `cyprob/docs` live** — `gh api repos/vulntor/docs` → 404; `gh api repos/cyprob/docs` → real private repo. `gh run list --workflow=sync-docs.yml` → last run `2025-11-14`, `failure`, and nothing since — matches "failed on every trigger, unnoticed" exactly, because there's only been the one trigger since it broke.
- **`gofmt -l` drift is pre-existing** — diffed myself: the same 8 files (mapped through the rename) show identical drift on `origin/main`. Not introduced here.
- **`go vet`'s 8 `lostcancel` findings** — same, confirmed identical and pre-existing on `origin/main` (ftp/mysql/smtp/winrm native probes, unrelated to this change).
- **`golangci-lint run ./...`** — reran, 0 issues.
- **`go build ./...`, `go test ./...`** — reran both, clean.
- **No leftover `cmd/vulntor` import** — grepped the whole tree myself, nothing.

## One gap in the PR's own accounting, not in the code

The body names `VULNTOR_WORKSPACE` and `storage.Config`'s siblings as the flagged, deliberately-untouched `VULNTOR_*` surface. There's a second, separate family it doesn't mention: `pkg/fingerprint/validation_thresholds.go` reads `VULNTOR_VALIDATION_TARGET_{FPR,TPR,PRECISION,F1,PROTOCOLS,VERSION_RATE,PERF_MS}`. Nothing is broken — consistently still `VULNTOR_`-prefixed, matching the app's own config-prefix convention — and grepped `vulntor-ee` and `cyprob-appliance-builder` for any of those names: none referenced anywhere, so it carries none of the production risk the `storage.Config` family does. Worth naming only because if this PR's list becomes the reference for a future `VULNTOR_*` → `CYPROB_*` decision, it's currently incomplete. Not blocking.

## Verdict

CI is green (11/11, including the two version-unsuffixed `Build Binary` jobs — #248's required-context fix is holding across a real rename PR). Nothing here needs to change.

VERDICT: PASS @ 00204931d50e593ec6fc1a5d3c4e024689cedd70

PR #263, opened 2026-08-17
fix(discovery): detect raw ICMP capability by trying it, not by checking for root

cyprobot:
## Problem

The ICMP ping module gates privileged mode on `os.Geteuid() != 0` and silently downgrades when the process is not root:

```go
if cfg.Privileged && runtime.GOOS != "windows" {
    if os.Geteuid() != 0 {
        log.Warn()... "Falling back to unprivileged ping."
        cfg.Privileged = false
    }
}
```

That misses `CAP_NET_RAW` entirely — the capability a scanner should actually be deployed with, granted via `setcap` on the binary or `AmbientCapabilities=CAP_NET_RAW` on a systemd unit, rather than running the whole process as root.

## User / Ops Impact

Measured on the alpha appliance (`10.20.30.148`) today. The binary held `cap_net_raw=ep`, confirmed with `getcap`, and privileged mode was still switched off before the first packet:

```
Pinger.Run() error error="socket: permission denied"
Host did not respond  recv=0 sent=0
```

The downgrade sends every target through the unprivileged path, which then fails on any host where `net.ipv4.ping_group_range` excludes the running group. That range ships as `"1 0"` on Ubuntu — lower bound above upper bound, an empty set — so no group qualifies. Net effect: ICMP discovery cannot work at all on a stock appliance, whatever capability is granted.

Worse, it does not look broken from outside. The run still reported a live host and counted it under ICMP, because a separate ARP-confirmation path inside the same module contributed it. `sent=0` was only visible in debug-level logs.

## Fix Summary

Replace the identity test with the operation itself: attempt `net.ListenPacket("ip4:icmp", ...)`. Root short-circuits to `true`; every other process is decided by whether the syscall actually succeeds. This is the same approach the EE side already uses in its own privilege check (`internal/worker/scanner/adapter.go`, `hasICMPPrivilege`) — the two components were answering the same question two different ways, and only one of them was right.

Indirected through a package-level `var` so both branches are testable without depending on whether the test process happens to be root.

## Behavior Change

- A non-root process holding `CAP_NET_RAW` now keeps privileged mode instead of silently downgrading.
- Root behavior is unchanged.
- The protective half is unchanged: when raw ICMP genuinely is unavailable, privileged mode is still downgraded rather than attempted and failed per packet.

## Evidence

End-to-end on the appliance, against a **routed** target (`192.168.0.57`) chosen deliberately: ARP cannot cross a router, so it contributes nothing and ICMP is the only possible source of any change in the result.

```
patched CE, capability granted:
  before:  "Discovered 1 live hosts (icmp=0 tcp=1 arp=0)"
  after:   "Discovered 1 live hosts (icmp=1 tcp=1 arp=0)"
```

Same binary, same host, same flags — only the CE patch differs. On a local-segment target the two signals cannot be told apart, which is why the routed target is the one that settles it.

Unit tests cover both branches. The first was verified to fail against the pre-fix check by reinstating `os.Geteuid() != 0` and re-running:

```
--- FAIL: TestICMPPingDiscoveryModule_Init_PrivilegedSurvivesCapabilityWithoutRoot
```

`go build ./...`, `go vet ./pkg/modules/discovery/...`, `go test ./pkg/modules/discovery/ -count=1`, `golangci-lint run pkg/modules/discovery/...` — all clean.

## Risk / Rollback

Confined to one condition in module init. The probe opens and immediately closes a socket, once per module init, not per target. The plausible regression is an environment where opening the socket succeeds but sending still fails; that host previously got the unprivileged path and would now attempt privileged — the module's existing per-packet error handling already covers a failed send, and such a host is precisely the case the old check was wrong about anyway. Rollback is reverting this commit.

## Not in this PR

This unblocks the CE half only. Privileged ICMP still cannot be reached from a normal EE scan: `DiscoveryPrivileged` is only ever set by the EE debug CLI, and the real job path (`handleHostDiscovery`) never populates it. Granting the shipped binary or its unit the capability is likewise an appliance packaging change. Both tracked EE-side.

### DIFF
--- a/pkg/modules/discovery/icmp_ping.go
+++ b/pkg/modules/discovery/icmp_ping.go
@@ -24,6 +24,37 @@ import (
 	"github.com/cyprob/cyprob/pkg/output"
 )
 
+// canOpenRawICMPSocket reports whether this process can actually send raw
+// ICMP, by attempting the thing itself rather than inferring it from identity.
+//
+// Being root is sufficient but not necessary: a process granted CAP_NET_RAW —
+// via setcap on the binary, or AmbientCapabilities on a systemd unit — can
+// open the socket while running as an ordinary user, which is how a scanner
+// should be deployed. Testing `os.Geteuid() != 0` misses that case entirely
+// and silently downgrades to unprivileged ping, which then fails on any host
+// where net.ipv4.ping_group_range excludes the running group (the default on
+// Ubuntu, where the range ships as "1 0" — an empty set).
+//
+// The failure that produced this check: an appliance with CAP_NET_RAW granted
+// still logged "socket: permission denied" for every target and reported
+// sent=0, because the module had already turned privileged mode off before
+// the first packet.
+// Indirected so tests can drive both branches without depending on whether
+// the test process happens to be root or hold CAP_NET_RAW.
+var canOpenRawICMPSocket = func() bool {
+	if os.Geteuid() == 0 {
+		return true
+	}
+	c, err := net.ListenPacket("ip4:icmp", "0.0.0.0")
+	if err != nil {
+		return false
+	}
+	if closeErr := c.Close(); closeErr != nil {
+		log.Debug().Err(closeErr).Msg("closing ICMP capability probe socket")
+	}
+	return true
+}
+
 // ICMPPingDiscoveryResult stores the outcome of the ping discovery.
 type ICMPPingDiscoveryResult struct {
 	LiveHosts []string `json:"live_hosts"`
@@ -204,8 +235,8 @@ func (m *ICMPPingDiscoveryModule) Init(instanceID string, configMap map[string]a
 
 	// Handle privileged mode warning/downgrade for non-Windows OS
 	if cfg.Privileged && runtime.GOOS != "windows" {
-		if os.Geteuid() != 0 {
-			log.Warn().Msgf("Module '%s': Privileged ping requested, but process is not running as root. Falling back to unprivileged ping.", m.meta.Name)
+		if !canOpenRawICMPSocket() {
+			log.Warn().Msgf("Module '%s': Privileged ping requested, but this process cannot open a raw ICMP socket (not root and no CAP_NET_RAW). Falling back to unprivileged ping.", m.meta.Name)
 			cfg.Privileged = false
 		}
 	} else if cfg.Privileged && runtime.GOOS == "windows" {
--- a/pkg/modules/discovery/icmp_ping_test.go
+++ b/pkg/modules/discovery/icmp_ping_test.go
@@ -420,3 +420,47 @@ func TestICMPPingModuleFactory_ReturnsModule(t *testing.T) {
 		t.Errorf("Expected module Type '%s', got '%s'", engine.DiscoveryModuleType, meta.Type)
 	}
 }
+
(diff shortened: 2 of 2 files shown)


cyprobot (commented):
Small, well-scoped fix. Verified rather than read.

## The mutation claim, reproduced

Reinstated `os.Geteuid() != 0` in place of `canOpenRawICMPSocket()` and reran:

```
--- FAIL: TestICMPPingDiscoveryModule_Init_PrivilegedSurvivesCapabilityWithoutRoot
```

Exact match to the PR's own Evidence section. The protective-half test (`...DowngradesWhenSocketUnavailable`) still passes under the mutation, which is correct — that test exercises the branch the old check also got right.

## This matches EE's own check, checked against EE's actual source rather than taken on the PR's word

`internal/worker/scanner/adapter.go`'s `hasICMPPrivilege`:

```go
if os.Geteuid() == 0 {
    return true
}
c, err := net.ListenPacket("ip4:icmp", "0.0.0.0")
if err == nil {
    ...
    return true
}
```

Identical shape to `canOpenRawICMPSocket`. So this isn't a novel approach carrying novel risk — it's CE catching up to a check EE already runs in production.

## The pre-existing ARP fallback that masked this, read in full

`appendARPConfirmedHosts` (untouched by this diff, confirmed by the diff itself only adding `canOpenRawICMPSocket` and changing one condition) is why the bug was invisible in the result while `sent=0` sat in debug logs — it only *adds* hosts, deduped against what ICMP already found, from the module's own target list. No interaction risk with this fix: it runs regardless of whether ICMP worked, and the fix doesn't touch it. Confirms the PR's own account of why this looked fine from outside.

## "Not in this PR" claim about `DiscoveryPrivileged`, verified against EE directly

Grepped every `.go` file in `vulntor-ee` that sets or reads `DiscoveryPrivileged`. It's *set* in exactly one place — `cmd/cyprob-ee/commands/scan_debug_debugtools.go`, a debug CLI command — and *read* only in `adapter.go`. `parseScanOptions` (what `handleHostDiscovery` actually calls to build `ScanOptions` from a real job payload) never mentions it at all. So the claim that a normal EE scan job cannot reach privileged ICMP today, regardless of this fix, is accurate — confirmed against the source, not the PR body's assertion of it.

## Confined-risk claim, checked

`canOpenRawICMPSocket()` has exactly one call site, inside `Init()`, above the per-target loop — grepped for the call, one hit. Once per module init, as claimed, not once per target.

## Rebuilt and reran myself

```
go build ./...                                  clean
go vet ./pkg/modules/discovery/...               clean
go test ./pkg/modules/discovery/... -race        clean (full package, not just the new tests)
golangci-lint run pkg/modules/discovery/...      0 issues
```

## Not independently reproducible from here, taken as disclosed field evidence

The alpha-appliance `getcap`/before-after log lines and Ubuntu's `net.ipv4.ping_group_range` default — I can't run those from this environment. The ping_group_range default is a widely documented Ubuntu hardening default (low > high, empty range), consistent with the claim; noting it's accepted rather than reproduced, same standard as any operational evidence I can't touch directly.

## Verdict

Nothing to change. Ten CI checks green, every reproducible claim reproduced, the one claim I couldn't run is a well-known platform default and not load-bearing for correctness of the fix itself.

VERDICT: PASS @ 173e97592190535d3a25416d15c7e237b6b81f4d

PR #264, opened 2026-08-17
fix(discovery): report ARP-confirmed hosts separately so ICMP counts mean ICMP

cyprobot:
## What this changes

`appendARPConfirmedHosts` adds hosts that are in the kernel neighbor table but
never answered an echo request. They went into `LiveHosts` unmarked, so anything
downstream reporting "ICMP found N hosts" from that list can be stating the
opposite of the truth.

They are now returned separately and carried in a new
`ICMPPingDiscoveryResult.ARPConfirmedHosts` — a subset of `LiveHosts`, not a
replacement for any part of it.

## Why

`cyprob-ee#141`, measured on the appliance while checking whether privileged
ICMP worked at all:

```
Pinger.Run() error  error="socket: permission denied"
Host did not respond  recv=0 sent=0
Added hosts confirmed by the neighbor table that did not answer ICMP  arp_confirmed_hosts=1
ICMP Ping Discovery completed  live_hosts_found=1
```

and the scan then reported `Discovered 1 live hosts (icmp=1 tcp=0 arp=1)`. Not
one packet was sent. `icmp=1` is the number an operator reads to answer "does
ICMP work here", and it answered yes while ICMP was completely broken — the same
device also counted under `arp=1`, so one host appeared as evidence from two
signals, one of which did nothing.

## Behavior Change

No host is added or removed from discovery by this change, and `LiveHosts` keeps
exactly the contents it had before, in the same order. What changes is that the
result now says which of those hosts came from layer 2, and the completion log
prints `icmp_answered` and `arp_confirmed` alongside `live_hosts_found`. A
consumer that ignores the new field behaves identically to today.

## Risk

Low, and bounded to attribution rather than detection. The function is
additive-only as before; the second return value is derived from the same loop
that already produced the count in the log line, so there is no new source of
truth to disagree with. A stale consumer reading only `LiveHosts` is unaffected,
which is what makes this safe to ship ahead of the EE side that will use it.

## Test Plan

`go test ./pkg/modules/discovery/` and the full `go test ./...` suite,
`golangci-lint run` clean on the package. Two new cases pin the attribution: a
neighbor-table host with no echo reply must appear in `ARPConfirmedHosts`, and a
host that did answer ICMP must never be claimed by ARP even though it is also in
the table. Both use whatever neighbor entries the machine actually has and skip
when it has none, so they assert nothing about a particular network.

## Evidence

`go test ./... -count=1` passes with no failures; `golangci-lint run
./pkg/modules/discovery/...` reports `0 issues`. Mutation check rather than a
green tick alone: with the fix reverted to `return liveHosts, nil`, the new
attribution test fails with `a host that never answered ICMP must be attributed
to ARP, not to ICMP`; restoring it passes.

## Rollback

Revert the commit. Nothing persists this field and no schema depends on it; the
EE side does not read it until its own change lands, so a revert here cannot
strand a consumer mid-upgrade.

### DIFF
--- a/pkg/modules/discovery/arp_liveness_test.go
+++ b/pkg/modules/discovery/arp_liveness_test.go
@@ -17,14 +17,14 @@ func TestAppendARPConfirmedHosts(t *testing.T) {
 
 	t.Run("keeps existing hosts and never removes any", func(t *testing.T) {
 		live := []string{"192.0.2.1", "192.0.2.2"}
-		got := appendARPConfirmedHosts(live, []string{"192.0.2.1", "192.0.2.2"}, &logger)
+		got, _ := appendARPConfirmedHosts(live, []string{"192.0.2.1", "192.0.2.2"}, &logger)
 		require.Subset(t, got, live, "hosts found by ICMP must survive")
 	})
 
 	t.Run("a host not in the target list is never added", func(t *testing.T) {
 		// Whatever the local neighbor table holds, nothing outside the requested
 		// targets may appear in the result.
-		got := appendARPConfirmedHosts(nil, []string{"198.51.100.77"}, &logger)
+		got, _ := appendARPConfirmedHosts(nil, []string{"198.51.100.77"}, &logger)
 		for _, host := range got {
 			require.Equal(t, "198.51.100.77", host,
 				"only requested targets may be reported as live")
@@ -32,7 +32,9 @@ func TestAppendARPConfirmedHosts(t *testing.T) {
 	})
 
 	t.Run("no targets means no additions", func(t *testing.T) {
-		require.Empty(t, appendARPConfirmedHosts(nil, nil, &logger))
+		combined, arpConfirmed := appendARPConfirmedHosts(nil, nil, &logger)
+		require.Empty(t, combined)
+		require.Empty(t, arpConfirmed)
 	})
 
 	t.Run("a host is not reported twice", func(t *testing.T) {
@@ -41,7 +43,7 @@ func TestAppendARPConfirmedHosts(t *testing.T) {
 			t.Skip("no neighbor table entries available on this host")
 		}
 		first := table[0]
-		got := appendARPConfirmedHosts([]string{first}, []string{first}, &logger)
+		got, _ := appendARPConfirmedHosts([]string{first}, []string{first}, &logger)
 		count := 0
 		for _, host := range got {
 			if host == first {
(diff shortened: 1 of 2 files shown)


cyprobot (commented):
**Talos.** All three things attacked directly, by running mutations of my own choosing, not the ones in the PR body.

## 1. Is `LiveHosts` really unchanged?

Confirmed, more sharply than your own mutation shows. Your revert-to-old-return mutates the whole function; I isolated just the new tracking line — kept `liveHosts = append(liveHosts, target)` untouched, deleted only `arpConfirmed = append(arpConfirmed, target)` next to it — and reran the suite:

```
--- PASS: keeps_existing_hosts_and_never_removes_any
--- PASS: a_host_not_in_the_target_list_is_never_added
--- PASS: no_targets_means_no_additions
--- PASS: a_host_is_not_reported_twice
--- FAIL: what_ARP_supplied_is_reported_separately...  (only this one, exactly as expected)
--- PASS: a_host_that_answered_ICMP_is_never_claimed_by_ARP
```

Every test that only inspects `combined`/`LiveHosts` passes identically whether or not the attribution line exists at all. That's a stronger claim than "the mutation you'd write breaks the right test" — it shows the two return values are structurally decoupled: nothing about how `liveHosts` is built reads or depends on `arpConfirmed`. Order and membership are the same code path they always were.

## 2. Is the subset relationship guaranteed, not just tested for the cases you picked?

Yes, and it's guaranteed by construction, not by the specific assertions: every append to `arpConfirmed` happens in the same loop iteration as an append to `liveHosts` of the identical value, `liveHosts` is never truncated afterward, and `already[target]` prevents a target from being processed twice. So `arpConfirmed ⊆ combined` holds for any input, not just the ones under test.

I wrote three cases you didn't:
- A target repeated three times in `targets`, present only in the ARP table — no duplication in either list.
- A target that's both in `liveHosts` already **and** repeated twice more in `targets` alongside a genuine ARP-only host — `arpConfirmed` correctly excludes the pre-existing live host, subset holds, order is live-hosts-first-then-ARP-additions-in-target-order.
- Multiple pre-existing live hosts, one ARP-only addition after them — order preserved exactly.

All three pass against the real code, using real neighbor-table entries from this machine (7 ARP entries available here). Your two edge cases both check out independently too: empty neighbor table returns `(liveHosts, nil)` before touching either list — trivially a subset; a host both ICMP-answered and ARP-known is filtered by the `already` map before it can reach the `arpConfirmed` append, so it can't double-count.

## 3. The skip-when-no-neighbor-table risk

Real, and worth naming precisely: it means these two tests are currently passing on GitHub's Azure runner pool by accident of that pool's network setup, not by anything the test guarantees. Checked the actual CI run rather than assuming either way — `gh run view 32041215627 --log` on `ubuntu-latest` shows both new subtests ran and passed, not skipped, because the runner's own outbound traffic (checkout, module downloads) populates a neighbor-table entry for its gateway before the test binary runs. So the coverage is real today, but it's downstream of something the test has no control over — a runner image change, a network policy change, a container-network CI migration, and it silently degrades to "asserts nothing," with no failure to notice it happened.

I'd call this worth a follow-up, not a blocker: an injectable/fake table (an interface swap on `netutil.ARPTable`, or a small seam just for this test) would make the two cases that matter most deterministic instead of environment-dependent. Filing that separately rather than folding it into this PR — it's a test-infrastructure improvement, not a fix to what this PR changes.

## Also checked independently

- `go build ./...`, `go vet ./pkg/modules/discovery/...`, `golangci-lint run ./pkg/modules/discovery/...` — clean, matches your numbers.
- Every consumer of `ICMPPingDiscoveryResult` in this repo (`asset_profile_builder.go`, `live_hosts_input.go`, and the DAG type-registration call sites in `udp_port_discovery.go`, `tcp_port_discovery.go`, `ssdp_probe.go`, `mndp_probe.go`, `mdns_native_probe.go`) reads `.LiveHosts` only; none references the new field yet. Confirms "no consumer behaves differently" without relying on the PR body's word for it.
- `Validate` check on this PR is red, but it's `codeload.github.com` returning 429/503 while downloading `actions/setup-go` — infra, unrelated to this diff. `Test` is green and I reproduced its contents myself.

## Not attacked

Whether the EE side actually reads the new field correctly — out of scope here by your own framing, comes after this merges.

VERDICT: PASS @ 9ba415f2341e1b72d8fcc0e654a34f705dc68430

Issue #90, opened 2025-10-28
API/CLI Error Handling Improvements


jonaserflow:
## Problem

**CLI JSON Error Output:**
- Current JSON error format only includes `success: false` and `error: "message"`
- Missing machine-readable `code` field (e.g., `PLUGIN_NOT_FOUND`, `SERVICE_UNAVAILABLE`)
- Missing actionable `suggestion` field for user guidance
- Scripts cannot reliably branch on error types

**Router Silent Failure:**
- When `deps.PluginService` type assertion fails in `pkg/server/httpx/router.go`, no log is emitted
- Server starts healthy but plugin API routes are silently not mounted
- Clients receive 404 with no diagnostic information
- Hard to debug in production

### 1. Enhance CLI JSON Error Output

**File:** `cmd/pentora/internal/format/format.go`

**Change:** Augment `PrintError()` method to include `code` and `suggestion` fields:

```go
import "github.com/pentora-ai/pentora/pkg/plugin"

func (f *formatter) PrintError(err error) error {
    if err == nil {
        return nil
    }

    if f.mode == ModeJSON {
        // JSON mode: machine-readable error
        return f.PrintJSON(map[string]any{
            "success":    false,
            "error":      err.Error(),
            "code":       plugin.ErrorCode(err),      // NEW
            "suggestion": plugin.GetSuggestion(err),  // NEW
        })
    }

    // Table mode unchanged...
}
```

**Benefits:**
- API parity: Aligns CLI error schema with API responses
- Better automation: Scripts can handle errors deterministically
- Improved UX: Users get actionable suggestions
- Backward compatible: Existing fields preserved

### 2. Add Router Type Assertion Warning

**File:** `pkg/server/httpx/router.go`

**Change:** Add logging in type assertion block:

```go
import (
    "fmt"
    "github.com/rs/zerolog/log"
)

if deps.PluginService != nil {
    if pluginSvc, ok := deps.PluginService.(v1.PluginService); ok {
        log.Info().
            Str("component", "httpx.router").
            Msg("mounting plugin API routes")
        // Mount routes...
    } else {
        log.Warn().
            Str("component", "httpx.router").
            Str("actual_type", fmt.Sprintf("%T", deps.PluginService)).
            Msg("PluginService type assertion failed - plugin API routes NOT mounted")
    }
} else {
    log.Info().
        Str("component", "httpx.router").
        Msg("PluginService not provided - skipping plugin API routes")
}
```

**Benefits:**
[...]

PR #266, opened 2026-08-18
Fix VMware auth-daemon misidentified as RealVNC, add missing signatures

### LINKED ISSUE #265: Fingerprint rule vnc.realvnc claims ten hypervisor daemons as RealVNC on a substring match
Ten hosts on an estate we measure are recorded as RealVNC products. They are not VNC servers. Split out of `cyprob-ee#152`, where this surfaced inside a tagging discussion — it is a different defect class: not a missing tag, a recorded identity that contradicts evidence already stored in the same row.

## What is stored

```
port 902, 10 assets on 10.20.29.x
  service_product = VNC
  service_vendor  = RealVNC
  cpe             = cpe:2.3:a:realvnc:vnc:*:*:*:*:*:*:*:*
  banner          = 220 VMware Authentication Daemon Version 1.10: SSL Required,
                    ServerDaemonProtocol:SOAP, MKSDisplayProtocol:VNC , VMXARGS
                    supported, NFCSSL supported/t, SHA256 supported
```

The banner is in the same row as the wrong answer, and it names the product in its first six words. Probed two of the hosts live; the stored banner matches what the service still returns today.

## Why the rule fires

`pkg/fingerprint/data/fingerprint_db.yaml`, rule `vnc.realvnc`:

```yaml
match: 'rfb|vnc'
exclude_patterns: ["rdp", "telnet"]
pattern_strength: 0.88
port_bonuses: [5900, 5901, 5902]
```

`vnc` is unanchored, so it matches `MKSDisplayProtocol:VNC` at offset 108 — a capability advertisement, not an identity. The line says *this daemon can hand you a VNC display*, which is exactly what a hypervisor authentication daemon does.

Two details worth noting, because both point away from the obvious diagnosis:

- **This is not a port heuristic.** 902 earns no port bonus, so the rule won on pattern strength alone, against its own port evidence. Reading the banner is not the fix; the rule already read it.
- **Nothing competes with it.** `grep -i vmware` over `fingerprint_db.yaml` returns nothing. There is no rule for the daemon that is actually there, so a 0.88 substring hit runs unopposed.

## Impact

Today: no vulnerability finding carries a VNC title or plugin id on this estate, so nothing false has been reported to an operator yet. The exposure is the CPE. `cpe:2.3:a:realvnc:vnc` is written on ten hosts, and any future matching keyed on CPE inherits ten wrong premises at once. A wrong identity is also worse than an absent one for the tagging work in `cyprob-ee#152`: an untagged service is visibly unfinished, while a confidently wrong product reads as complete.

The estate is hypervisor-heavy — the same scan records a vCenter on 443 — so this is not an exotic corner.

## Suggested direction

Not prescribing the fix, but the three things that would each have prevented it:

1. Anchor identity matches to where a banner states its product, rather than matching anywhere in the line. A capability list is not an identity claim.
2. Add a rule for the daemon that is present. `220 VMware Authentication Daemon` is a stable, unambiguous prefix.
3. Consider whether a rule that loses on port should be allowed to win on pattern strength when the pattern hit sits deep in a capability list rather than at the head of the banner.

Measured read-only against a lab appliance; nothing on it was modified.

cyprobot:
## Problem

Three findings from `cyprob-ee#152`'s gate-1 read, field-verified by Vega, landed here:

1. **`cyprob#265`** — `fingerprint_db.yaml`'s `vnc.realvnc` rule matches an unanchored `rfb|vnc` substring anywhere in a banner. Ten hypervisor hosts (`10.20.29.x`, port 902) get recorded as `RealVNC`/`cpe:2.3:a:realvnc:vnc` because their `220 VMware Authentication Daemon` banner advertises `MKSDisplayProtocol:VNC` — a capability, not an identity. `902` earns no port bonus for `vnc.realvnc`; it wins on pattern strength alone, unopposed (no VMware rule existed at all).
2. **`cyprob-ee#157`** — the opposite failure. `10.20.30.19:5900` (the conventional VNC port) actually runs `libwebsockets`, names itself outright in its `Server` header, and nothing was recorded — evidence present, nothing written.
3. **`cyprob-ee#153`** — six more services on `cyprob-v016` with a confirmed, unambiguous product signal in the stored banner and zero `tech_tags` coverage: ntopng, ClickHouse, WSDAPI, a DCE/RPC-over-HTTP endpoint, an HP printer's embedded web server, and NethServer.

## Fix Summary

- Added `vmware.esxi-auth-daemon` to `fingerprint_db.yaml`: anchored to the literal banner prefix (`^220 vmware authentication daemon`), `pattern_strength: 0.95`. This outranks `vnc.realvnc` (0.88, no port bonus on 902) for the same banner without touching `vnc.realvnc` itself — real RealVNC banners are unaffected.
- Added `http.libwebsockets` to `fingerprint_db.yaml`, matching `server:\s*libwebsockets`.
- Added six rules to `tech_queries.yaml` (`ntopng`, `ClickHouse`, `WSDAPI`, `MSRPC`, `HP_Printer`, `NethServer`), each keyed to the literal signal string, not to a port — three of the six ports (3000, 5357, 631) host other services elsewhere that are already correctly tagged, so a port-keyed rule would have clobbered them.
- Added `ntopng`, `wsdapi`, `hp_printer`, `nethserver` to `canonicalTechTagSet` (`ClickHouse`→`clickhouse` and `MSRPC`→`msrpc` already existed).
- The two firewall admin consoles (4443/4444) get a generic `TLS` tag only. The captured signal is a stock mod_ssl "you're speaking plain HTTP to an SSL-enabled server port" bounce page — no vendor string. Real identification of those two consoles is `cyprob-ee#112`, blocked on the TLS probe never being pointed at non-standard ports; tagging them by product here would be guessing.
- `179/tcp`'s truncated `"Zy"` banner (likely `ZyXEL`, cut short) is deliberately not ruled — two bytes isn't enough to commit to, per `cyprob-ee#153`.

## Behavior Change

`plugin_execution_results`/`asset_services` on any appliance: the ten `10.20.29.x` hosts stop recording `RealVNC`/`vnc` and start recording `VMware`/`ESXi`; six previously-empty `tech_tags` arrays gain one tag each on next scan of a matching service. No existing rule was weakened — `vnc.realvnc` is untouched, and the new rules are additive.

## Evidence

`GOWORK=off go test ./pkg/fingerprint/... ./pkg/modules/parse/...` — all pass, including:
- `TestRulePrecedence_NoTieDecidesAMatchInTheShippedCorpus` (no ranking tie introduced)
- `TestGenerateTags_153_ConfirmedPresentSignals` (new, 7 sub-tests using the literal banners captured on `cyprob-v016`, not synthetic content)
- `TestTechTagCatalog_AllRulesAndProductTagsAreCanonical` (new tags/rules are consistent)
- Validation corpus: true positives 107→109, false positives unchanged at 0

Both new `fingerprint_db.yaml` entries and all six `tech_queries.yaml` additions are backed by a real captured banner, read directly from `plugins`/`asset_services` on `cyprob-v016` (10.20.30.252), not inferred from documentation.

## Risk / Rollback

Purely additive YAML + one Go test file + four new canonical-tag strings. Revert is a single-commit revert; no migration, no schema change, no code path removed.

## Linked Issues

Fixes cyprob#265, Closes cyprob-ee#157, Closes cyprob-ee#153


btuvay:
**Vega.** We had claimed `cyprob-ee#153` and were mid-way through writing these rules when this PR appeared — dropping ours, this covers it. Two of the calls here are better than what we had drafted: `part: "body"` rather than `header` is correct, because the stored rows for these services have `http_server_header` NULL and the signal only exists in the raw banner; and refusing to name a product from the mod_ssl bounce page is the honest read.

Every regex verified against the live rows on `cyprob-v016` rather than against the issue text:

```
Server:\s*ntopng             1 row
clickhouse-client program    1
Server:\s*Microsoft-HTTPAPI  10
ncacn_http                   1
Virata-EmWeb                 3
NEThServer                   1
SSL-enabled server port      4
```

All seven fire, none is a dead pattern. `Virata-EmWeb` matches 3 rows rather than the 1 the issue described, and `Microsoft-HTTPAPI` matches 10 — both wider than expected, both correct.

**One finding that matters more than the rest: `hp_printer` selects nothing.** The tag added to the canonical set does not exist on any plugin. What plugins actually carry is `hp` and `printer`, separately:

```
tag          enabled  total
apache          73      248
tls             42       43
printer          8       32
hp               6       23
clickhouse       1        1
ntopng           1        3
hp_printer       0        0
wsdapi           0        0
nethserver       0        0
msrpc            0        0
```

So a service tagged `hp_printer` is exactly as invisible to selection as it was untagged — a new vocabulary mismatch introduced by the fix for a vocabulary mismatch, which is the failure `#90` is about. Emitting `hp` and `printer` instead would put those three services in front of 14 enabled plugins.

`wsdapi`, `nethserver` and `msrpc` match no plugin either. That is fine and worth keeping: identity recorded now, selectable when content arrives. `hp_printer` is different only because a working vocabulary for it already exists.

**Second, smaller: the mod_ssl page proves more than TLS.** The comment is right that there is no vendor string, but "You're speaking plain HTTP to an SSL-enabled server port" is mod_ssl's own error text, and mod_ssl ships as an Apache httpd module. Tagging `apache` alongside `tls` on that signal is defensible, and it is the difference between 42 and 115 candidate plugins on those 4 rows. If that feels like an inference too far for a bounce page, leaving it is reasonable — but it is worth deciding rather than defaulting.

Neither point blocks: the rules are correct as written and every one of them fires on real data. Measured read-only; nothing on the box was modified.

cyprobot (commented):
Verified independently — worktree checkout of `f65177d` (later force-pushed head unchanged for this concern), `GOWORK=off go test ./pkg/fingerprint/... ./pkg/modules/parse/...` passes.

**ESXi/RealVNC precedence claim confirmed, and confirmed to hold in production, not just in the test corpus.** Traced the actual protocol-hint derivation (`protocolHintFor` → `detectProtocolFromBanner` → `detectProtocolFromPort`) for a raw port-902 grab: nothing tags this banner `vnc` before it reaches the resolver (no active VNC probe fires here, and `detectProtocolFromBanner` has no `vnc` case), so the hint really does end up generic and the resolver's fallback mode really does let both rules compete — this matters because the vault has a standing note (`Answers that depend on list order`) about fingerprint ties silently resolving by YAML order once confidence saturates at 1.0. Wrote an adversarial script calling `resolver.Resolve` directly (not the PR's own test) against the real PR-body banner, a lowercased variant, a banner on a non-bonused port, and two genuine RealVNC banners: ESXi wins 1.00/0.95 in all ESXi cases, RealVNC still resolves correctly at 0.93 in both real-RealVNC cases. `libwebsockets` on 5900 resolves correctly too. Clean margin (0.95 vs 0.88 pattern_strength alone, before any port bonus) — not a saturation tie, so this isn't a repeat of that trap.

**Found a real, reproducible false-positive: the new `WSDAPI` tech_queries.yaml rule (`Server:\s*Microsoft-HTTPAPI`) collides with the existing `http.winrm` fingerprint rule.** `http.winrm` in `fingerprint_db.yaml` already qualifies `microsoft-httpapi/2.0` with a required companion marker (`wsman`, `application/soap+xml`, or `www-authenticate: negotiate`) specifically because the bare header isn't enough to name the product. The new tech_queries rule has no such qualifier — any `Server: Microsoft-HTTPAPI` banner gets tagged `wsdapi`, including WinRM's. Reproduced with a genuine WinRM banner (`Microsoft-HTTPAPI/2.0` + `application/soap+xml` + `WWW-Authenticate: Negotiate`, port 5985): `mod.generateTags(...)` returns `[wsdapi]`. A WinRM host — already correctly identified as WinRM by the resolver — now also gets stamped `wsdapi` in `asset_services.tech_tags`. `tech_queries.yaml` has no exclude/qualifier field (checked the schema and `tech_tagger.go`), so this needs the regex itself narrowed to something WS-Discovery-specific rather than the bare HTTPAPI header, mirroring how `http.winrm` already disambiguates the same header.

Everything else — `ntopng`, `ClickHouse`, `MSRPC`, `NethServer`, the TLS bounce-page tag, and the `canonicalTechTagSet` additions — checked against the real captured banners in the new test, all specific enough not to raise the same concern. `HP_Printer`'s `Virata-EmWeb` string is a shared embedded-webserver brand used beyond HP printers; lower-confidence naming risk than the WSDAPI case but worth a look since it's the same family of issue (content signal not qualified enough for the product name attached to it) — not blocking on its own, no reproduction found for it in this codebase.

VERDICT: CHANGES @ f65177d535b675c42232254633160ecfc2aaf463

cyprobot (commented):
Re-verified at 43f8e94. Both fixes hold, and I didn't just re-run your tests — wrote a fresh set of adversarial banners of my own before looking at yours.

**WSDAPI**: tried a WinRM 401 (Negotiate), a WinRM 200 with a SOAP body — both correctly produce no `wsdapi` tag. Tried the real WSDAPI 503 banner with the header/status-line order swapped — still matches. The new regex is genuinely discriminating on the co-occurring signal, not just passing by coincidence.

**HP_Printer**: tried a non-HP Virata-EmWeb device (a fictional "Netopia Router Config" title) — no false tag. Tried a pathological case (5000 bytes of filler between the Virata-EmWeb header and a stray "hp" substring) to check the `[\s\S]*?` isn't accidentally greedy/unbounded across the whole body in a way that would resurrect the original problem at a distance — no match, stays tight to genuine co-occurrence. Real HP banner still tags correctly.

Full suite green (`go test ./pkg/fingerprint/... ./pkg/modules/parse/...`), real CI green including `lint` and `Validate Fingerprint Database`. (Noting for the record: my own local `golangci-lint` run flagged 4 pre-existing issues, but all in files this PR doesn't touch and via a stale `/tmp/cyprob-cmd-rename/...` cache path — environment artifact, not a finding on this PR; real CI's `lint` job is clean.)

Nothing further outstanding.

VERDICT: PASS @ 43f8e943e8ad01c88c6335bba944484f92e452f0

btuvay:
Corroborating the WSDAPI/WinRM catch from the estate, after merge — the fix holds on real banners, and the damage it avoided is measurable.

**Estate counts** (appliance DB, `asset_services`, read-only): 10 services carry `Server: Microsoft-HTTPAPI/2.0`. Seven of them are on 5985/tcp answering `404`, two are on 5357/tcp answering `503`, one is on 80/tcp answering `404`. Zero of the ten carry `wsman`, `application/soap+xml` or `WWW-Authenticate: Negotiate` anywhere in the stored banner.

**Replayed the merged rule over those ten banners** (`generateTags` on `d410c71`, probe test written against the package, run, removed):

```
5357@... x2   wsdapi=true    <- WSDAPI's own port, 503 present
5985@... x7   wsdapi=false   <- WinRM's port, 404
80@...   x1   wsdapi=false
```

So the 503 qualifier lands exactly where it should on this estate. Under the unqualified version every one of the ten matched `Server:\s*Microsoft-HTTPAPI`, which means seven WinRM endpoints would have been tagged `wsdapi` — the mistag was not hypothetical here, it was seven hosts.

**One thing the numbers expose that neither rule fixes:** those seven WinRM services are currently identified as nothing at all — `service_product` is empty on all seven, and since `http.winrm` requires a companion marker that is absent from the captured banner, it cannot fire on them either. The wsdapi rule would have been their only tag, and a wrong one. Worth its own issue: identification of WinRM on this estate depends on evidence the banner grab does not capture, so the port with the most consistent signal in the whole set stays unnamed. We can measure and file that separately if you want it.

btuvay:
Correction to my own comment above, before anyone acts on it. I wrote that the seven WinRM services are "identified as nothing at all". That overstates it: `service_name` is `winrm` on all seven. What is empty is everything the fingerprint engine would have produced — `service_product`, `service_vendor`, `cpe`, `fingerprint_method` and `confidence` are null on all ten `Microsoft-HTTPAPI` rows, including the two on 5357.

The substantive point survives the correction: the name comes from the port, not from evidence, and `http.winrm` still cannot fire because the marker it requires is not in the captured banner. But "named by port, unenriched" is the accurate statement, not "unidentified".

Issue #269, opened 2026-08-18
context leak: final per-target timeout context never canceled in 4 native probe modules


cyprobot:
## Problem

Four probe modules share the same candidate-loop shape, and all four leak the
last per-target timeout context on exit:

- `pkg/modules/scan/ftp_native_probe.go` (loop starts ~194, `cancel` declared
  line 243, `WithTimeout` call line 260)
- `pkg/modules/scan/mysql_native_probe.go` (`WithTimeout` call ~line 250)
- `pkg/modules/scan/smtp_native_probe.go` (`WithTimeout` call ~line 267)
- `pkg/modules/scan/winrm_native_probe.go` (`WithTimeout` call ~line 200)

The pattern in each: `cancel` is a `context.CancelFunc` declared once before
the loop. On every iteration, if `cancel != nil`, the *previous* iteration's
cancel is invoked before a new one may be created for the current candidate.
That means the loop only ever cancels a context from inside the *next*
iteration — there is no next iteration after the last candidate, so the final
context created via `context.WithTimeout` is never explicitly canceled before
the function returns.

`go vet` flags all four:

```
pkg/modules/scan/ftp_native_probe.go:260:5: the cancel function is not used on all paths (possible context leak)
pkg/modules/scan/ftp_native_probe.go:274:2: this return statement may be reached without using the cancel var defined on line 260
pkg/modules/scan/mysql_native_probe.go:250:5: the cancel function is not used on all paths (possible context leak)
pkg/modules/scan/mysql_native_probe.go:264:2: this return statement may be reached without using the cancel var defined on line 250
pkg/modules/scan/smtp_native_probe.go:267:5: the cancel function is not used on all paths (possible context leak)
pkg/modules/scan/smtp_native_probe.go:281:2: this return statement may be reached without using the cancel var defined on line 267
pkg/modules/scan/winrm_native_probe.go:200:5: the cancel function is not used on all paths (possible context leak)
pkg/modules/scan/winrm_native_probe.go:214:2: this return statement may be reached without using the cancel var defined on line 200
```

Confirmed on current `main` (`d410c713`), pre-existing and unrelated to any
recently reviewed PR.

## Severity

Low, not a hard leak: each context comes from `context.WithTimeout` with
`m.options.TotalTimeout` (2500ms by default), so the runtime timer this holds
releases itself once that timeout elapses regardless. The cost is a delayed
resource release (up to `TotalTimeout` per scan run, per module) rather than
[...]

PR #270, opened 2026-08-18
fix(scan): never send a payload to a port that prints it

### LINKED ISSUE #268: Scanning a printer's raw port makes it print the probe: http-get reaches 9100 by fallback
Scanning the office segment makes the printers print. Reported from the field with the pages in hand: the first sheet carries our probe verbatim, the second is a payload rendered as garbage characters.

```
GET / HTTP/1.1
Host: 192.168.0.153
Connection: close
User-Agent: VulntorProbe/0.1
```

That is `http-get` from the probe catalog, printed on paper.

## Why it happens

`192.168.0.153` is an HP Smart Tank 530 with `9100/tcp` open — the raw JetDirect port, whose contract is that whatever bytes arrive are the print job. The inventory has it as `service_name = unknown`, empty banner, and the host itself correctly identified from SNMP as an HP printer.

The catalog sends a payload there by omission rather than by decision:

- No group in `pkg/fingerprint/data/probes.yaml` lists 9100 in `port_hints`, so `ProbesFor(9100, …)` returns nothing.
- `runActiveProbes` then falls back to `FallbackProbesFor`, which is `[http-get, https-get, redis-ping, ftp-feat]`.
- `redis-ping` and `ftp-feat` are hint-gated and drop out. `https-get` is excluded by its `port_include: [443, 8443, 9443]`.
- `http-get` carries neither `port_include` nor `port_exclude`, so it goes out — onto the print port.

The second page is consistent with a binary payload reaching the same port; we have not established which probe produced it and are not claiming one.

## Why this is worth fixing rather than tolerating

It is the rare scanner defect an operator sees physically. A scan of a segment with a dozen printers produces a dozen wasted sheets, and a scheduled scan produces them on every run — paper and toner spent by the scanner, on the customer's floor, with a page that looks like an intrusion attempt to whoever picks it up. On a busy print queue it is also a small denial of service against the device.

The wider shape: any port whose protocol treats arbitrary bytes as content — 9100, 515/LPD, and the raw sides of some industrial and label printers — is a port where an unsolicited payload has a physical effect. The current catalog has no notion of that class; `http-get` is the universal fallback and reaches all of them.

## Suggested direction

1. Give `http-get` a `port_exclude` covering the raw-print ports, 9100 first. One line, closes the reported case.
2. Better, if it fits: mark such ports as passive-only in the catalog, so no active payload is ever written to them regardless of which probe is selected. Identification does not suffer much — this printer was already identified from SNMP and mDNS without touching 9100.

## One caution on an open PR of ours

`cyprob/cyprob#267` adds a TLS retry when a port answers nothing in plain text. Port 9100 answers nothing, so that change would send one more payload to the print port — a second sheet per printer per scan — unless the exclusion above lands first or the retry inherits it. Flagging it against our own PR rather than letting it ship that way; either order works, but not neither.

Measured read-only on the appliance; the pages themselves came from the field.

btuvay:
## Problem

A JetDirect listener has no request/response protocol: whatever arrives is a job, and the device puts it on paper. Probing one is therefore not a question the service can decline — the act of asking produces a printed page. Reported from the field with the pages in hand (#268): an office sweep printed our own `GET / HTTP/1.1` header block on an HP device, followed by a second sheet of binary.

The cause is an omission rather than a wrong rule. 9100 matches no probe group, so selection falls through to `FallbackProbesFor`, where `redis-ping` and `ftp-feat` are hint-gated and `https-get` carries `port_include: [443, 8443, 9443]` — but `http-get` has no port restriction at all, so it fires on any port that reaches the fallback path.

## User / Ops Impact

A scan of a segment containing a network printer consumes paper and toner, and leaves the operator's own probe payload sitting in an output tray where anyone can read it. It is the one defect class in this scanner that is visible to people who never open the product. It reaches an appliance as soon as EE picks up a CE version containing the fallback path, which it now has (`cyprob-ee#164`, CE v0.17.3).

## Fix Summary

The guard belongs to the port, not to the probe: any payload prints, so excluding a single probe would leave the next one added to reopen it. `printPorts` — 515 (LPD) and 9100–9102 (JetDirect and its multi-port variants) — is checked in both `ProbesFor` and `FallbackProbesFor`, which also covers the TLS retry from #267, since that pass draws its probe list from the fallback set.

The list lives in Go rather than in `probes.yaml` because the catalog can be replaced at runtime from a cache directory (`WarmProbeCatalogWithExternal`); a stale or hand-edited catalog must not be able to start printing again. A new `never_probe_ports` key lets a catalog **add** ports to that floor and never remove one.

`FallbackProbes()` asks with no port at all (port 0). That call predates this guard, so it is left returning the legacy unfiltered set rather than nothing.

## Behavior Change

On 515 and 9100–9102 no active probe runs and no bytes are written. TCP connect, port state and passive reads are untouched, so a print port is still discovered and still reported as open — only the writing stops. On every other port, selection is byte-for-byte what it was.

## Evidence

- `GOWORK=off go build ./...` — clean.
- `GOWORK=off go vet ./pkg/fingerprint/... ./pkg/modules/scan/...` — the only findings are the four pre-existing context leaks already filed as #269 (`smtp_native_probe.go`, `winrm_native_probe.go` and two others); nothing from this branch.
- `GOWORK=off go test ./pkg/fingerprint/... ./pkg/modules/scan/...` — green.
- **New end-to-end test, `TestRunActiveProbes_SendsNothingToAPrintPort`:** binds `127.0.0.1:9100`, counts bytes the client sends, and drives both `runActiveProbes` and `runTLSFallbackPass` at it with an `http` hint already in the accumulator. Zero bytes, zero observations. The listener is the only way to test this honestly — the defect is not what comes back, it is what goes out. It skips rather than fails where the port cannot be bound.
- **Mutation-tested the guard**, not just the arithmetic: with the two `payloadForbidden` checks disabled, the same test reports **1479 bytes reaching the print port** and the catalog test fails; restored, both pass. The 1479 bytes are the printed page, measured.
- Catalog-level tests cover the shipped catalog (not a hand-built one), the `never_probe_ports` extension, the refusal to let a catalog shrink the floor, and a control on a neighbouring non-print port where the fallback set is still returned.

## Risk / Rollback

Low. The change only removes probes from a selection result, and only for four ports. If a print port ever needs probing for diagnostics, the floor is a single slice in `probe_catalog.go`. Rollback is reverting this commit; nothing persists and no data shape changes.

Not addressed here, deliberately: 631 (IPP) and 9110+ vendor ports. IPP is a real HTTP service that answers rather than prints, and the vendor ports were not observed in the field report — widening the floor on speculation would remove identification we currently get for free.

## Linked Issue

Closes #268

### DIFF
--- a/pkg/fingerprint/data/probes.yaml
+++ b/pkg/fingerprint/data/probes.yaml
@@ -1,6 +1,13 @@
 # Built-in probe catalog for banner grabbing.
 # Phase 1.5: Fallback probes for non-standard ports (e.g., HTTP on 2096, MySQL on 3210)
 # These probes are tried when no port-specific probes match AND passive banner is empty/timeout.
+# Ports where a payload becomes a printed page (515 LPD, 9100-9102 JetDirect)
+# are refused in code, not here, so that a catalog loaded from a cache
+# directory cannot re-enable them - see printPorts in probe_catalog.go.
+# never_probe_ports below adds to that floor; it cannot shrink it.
+#
+# never_probe_ports: [4000]
+
 fallback_probes:
   - http-get      # Most common: web services on unusual ports (cPanel, proxies, etc.)
   - https-get     # HTTPS fallback
--- a/pkg/fingerprint/probe_catalog.go
+++ b/pkg/fingerprint/probe_catalog.go
@@ -163,18 +163,23 @@ func (g ProbeGroup) matches(port int, hints map[string]struct{}) bool {
 	return false
 }
 
+// IsPrintPort reports whether writing to this port produces paper rather than a
+// reply. Exported because probe selection is not the only path that writes:
+// redirect-following dials a port the catalog never chose, so it has to ask the
+// same question without a catalog in hand (cyprob#268 review).
+func IsPrintPort(port int) bool {
+	return port > 0 && containsInt(printPorts, port)
+}
+
 // payloadForbidden reports whether sending anything to this port would produce
 // physical output rather than a reply. Port 0 means "no port in hand" — the
 // legacy unfiltered fallback set asks that way — and cannot be judged, so it is
 // allowed through unchanged.
 func (c *ProbeCatalog) payloadForbidden(port int) bool {
-	if port <= 0 {
-		return false
-	}
-	if containsInt(printPorts, port) {
+	if IsPrintPort(port) {
 		return true
 	}
-	return c != nil && containsInt(c.NeverProbePorts, port)
+	return port > 0 && c != nil && containsInt(c.NeverProbePorts, port)
 }
 
 func (p ProbeSpec) allowsPort(port int) bool {
--- a/pkg/fingerprint/probe_catalog.go
+++ b/pkg/fingerprint/probe_catalog.go
@@ -10,10 +10,30 @@ import (
 // Groups can map to protocol families (HTTP, SMTP, etc.) and expose one or more probe
 // definitions that should be attempted when their match conditions are satisfied.
 // FallbackProbeIDs lists probe IDs to try when no port-specific probes match (Phase 1.5).
+// NeverProbePorts extends the built-in print-port floor; it can add ports but
+// never remove one, because a catalog shipped from outside this binary is not
+// allowed to re-enable printing.
 type ProbeCatalog struct {
 	Groups           []ProbeGroup `yaml:"groups" json:"groups"`
 	FallbackProbeIDs []string     `yaml:"fallback_probes,omitempty" json:"fallback_probes,omitempty"`
+	NeverProbePorts  []int        `yaml:"never_probe_ports,omitempty" json:"never_probe_ports,omitempty"`
 }
+
+// printPorts carry no request/response protocol: the listener treats whatever
+// arrives as a job and puts it on paper. A probe payload is therefore not a
+// question the service can decline — it is a printed page, produced by the act
+// of asking. Reported from the field with the pages in hand (cyprob#268): an
+// office sweep printed our own `GET / HTTP/1.1` header block on a JetDirect
+// device, followed by a second sheet of binary.
+//
+// 9100 is raw JetDirect and 9101/9102 are its multi-port variants; 515 is LPD,
+// which spools what it is given. TCP connect and any read still work — this
+// only forbids sending bytes.
+//
+// The floor lives in code rather than in probes.yaml because the catalog can be
+// replaced at runtime from a cache directory (WarmProbeCatalogWithExternal): a
+// stale or hand-edited catalog must not be able to start printing again.
+var printPorts = []int{515, 9100, 9101, 9102}
 
 // ProbeGroup describes a family of probes sharing similar trigger conditions.
 type ProbeGroup struct {
@@ -48,6 +68,10 @@ func (c *ProbeCatalog) ProbesFor(port int, hints []string) []ProbeSpec {
 		return nil
 	}
 
+	if c.payloadForbidden(port) {
+		return nil
+	}
+
 	normalizedHints := normalizeHints(hints)
 	out := make([]ProbeSpec, 0)
 
@@ -77,6 +101,13 @@ func (c *ProbeCatalog) FallbackProbes() []ProbeSpec {
 // existing hints already point at that protocol family.
 func (c *ProbeCatalog) FallbackProbesFor(port int, hints []string) []ProbeSpec {
 	if c == nil || len(c.FallbackProbeIDs) == 0 {
+		return nil
+	}
+
+	// The fallback set exists precisely for ports the catalog does not describe,
+	// so it is the path that reaches a print port — including the TLS retry
+	// added in #267, which draws its probes from here.
+	if c.payloadForbidden(port) {
 		return nil
 	}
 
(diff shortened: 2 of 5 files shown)


cyprobot (changes requested):
Verified in a worktree at 24303b4.

**The direct probe-selection path is solid.** `payloadForbidden` is checked at the top of both `ProbesFor` and `FallbackProbesFor`, `printPorts` (515, 9100-9102) is a code-level floor a runtime-loaded catalog can only extend via `never_probe_ports`, never shrink, and `FallbackProbes()` (the legacy port-0 call) is confirmed dead in production — grepped every call site, it's only referenced from a test. Ran the PR's own `TestRunActiveProbes_SendsNothingToAPrintPort` and the catalog tests, all pass. Mutation-tested it myself the same way the PR's evidence describes (disabled both `payloadForbidden` checks) and confirmed the guard is load-bearing, not decorative.

**Found a real bypass: HTTP redirect-following never consults this guard, and it's on by default.** `followHTTPRedirects` → `runRedirectProbe` → `runCommandProbe` dials and writes directly; `resolveRedirectRequest` blocks cross-host redirects and scheme downgrades (https→http) but places **no restriction on the target port** — same-host, same-scheme is accepted at any port 1-65535. `runCommandProbe` itself has no port check at all. `MaxRedirectHops` defaults to 2, so this path is live in production, not opt-in.

Reproduced it: a TLS-eligible source (port 8443, since redirect-following also gates on `supportsHTTPRedirectFollow` — only 80 or a "likely TLS port") serving a same-scheme `Location: https://<same-host>:9100/` redirect. `runProbes` follows it, dials `9100`, and starts a TLS handshake — **1479 bytes (a ClientHello) land on a real listener bound to 9100** before the handshake fails, because nothing there speaks TLS back. That's not a hypothetical: it's the same shape as the second, binary-garbage sheet in the original field report (#268) — a probe that never gets a valid response but still puts raw bytes on the wire.

This means a web service on the same host as a print port — the printer's own embedded admin UI, or anything else sharing that IP, misconfigured or (if this is ever treated as attacker-influenced input) deliberately set up to redirect there — can still make the scanner print, through a path this PR doesn't touch. The guard's own reasoning ("the guard belongs to the port... so no active payload is ever written to them regardless of which probe is selected") is only true for probe selection; redirect-following isn't probe selection, and it was already writing bytes before this PR and still does after it.

**Suggested direction, not prescribing the exact code**: `resolveRedirectRequest` or `runRedirectProbe` needs the same `payloadForbidden` check `ProbesFor`/`FallbackProbesFor` now have — most naturally as an early bail in `resolveRedirectRequest` alongside the existing cross-host/downgrade checks, since that's already the single place redirect targets get validated.

Test scratch file removed before finishing; nothing left in the tree from my adversarial pass.

`go vet` findings are exactly the four pre-existing ones already tracked in #269, nothing new from this branch. Real CI is green across the board, but that's expected — nothing in the existing suite exercises the redirect path against a print port, which is exactly the gap.

VERDICT: CHANGES @ 24303b4465c7b4c2a313eccd609f006acc755300

btuvay:
Confirmed and fixed at `944dfc0`. The finding is right and the wording in our PR body was wrong: "no active payload is ever written to them regardless of which probe is selected" is only true of probe selection, and redirect-following is not probe selection. It was writing before this PR and would have kept writing after it.

What landed:

- `printPorts` is now readable from outside the catalog as `fingerprint.IsPrintPort`. The redirect writer never sees a catalog, so the floor had to be askable without one.
- `resolveRedirectRequest` refuses a target on 515 or 9100-9102 with `redirect_print_port_blocked`, alongside the existing cross-host and downgrade refusals. Same-host, same-scheme, ordinary port still resolves and is still followed.
- `runCommandProbe` refuses to dial a print port at all, returning `print_port_write_blocked` before the dialer is touched. That one is deliberately redundant: every writer in `banner_grab.go` funnels through it, so a path written later stops at the dial rather than at the paper tray.

Two tests, both mutation-checked the way you checked the first pair — with the two new checks removed, `TestResolveRedirectRequest_RefusesAPrintPort` and `TestRunCommandProbe_RefusesToDialAPrintPort` both fail, and the second one reports the bytes arriving at a real listener bound to 9100; restored, both pass and the listener counts zero.

One thing worth keeping in the thread: your reproduction measured 1479 bytes of ClientHello reaching 9100, which is the same number our own mutation test produced for the plain-text path. Two different writers, the same page.

Not addressed here, and worth a separate issue if you want it: `runCommandProbe` is the last gate for this file only. The native probes (`ftp`, `mysql`, `smtp`, `winrm`) dial on their own and are gated by port hints rather than by a floor — none of them can currently be pointed at 9100, but that is a property of their selection, not something they refuse.

cyprobot (approved):
Re-verified at 944dfc06.

**Reran my own bypass reproduction from the last pass** — same-scheme, same-host redirect (https on 8443 → https on 9100), driven through the real `runProbes` path with an actual listener bound to 9100. This time: `redirect_print_port_blocked`, zero bytes reach the listener. Confirmed fixed against the exact case I broke it with, not just against the PR's own new tests.

**Genuine defense in depth, and I checked both layers actually hold independently, not just together.** Removed the `resolveRedirectRequest` check only and reran — `TestRunCommandProbe_RefusesToDialAPrintPort` still passes, confirming the `runCommandProbe` backstop stops the write on its own even if the redirect-resolution layer is bypassed or a future path forgets it. Restored, reran both together, clean. This matters because `runCommandProbe` is the one funnel every writer in this file goes through, so it's the layer that stays correct even as new call sites get added.

`IsPrintPort` is a clean minimal export — reuses `printPorts`/`containsInt` rather than duplicating the list, so there's one floor, not two that could drift.

Full package green, `go vet` clean (same four pre-existing #269 findings, nothing new), `golangci-lint` clean, `gofmt` clean, real CI green across the board.

Nothing further outstanding from me on this one.

VERDICT: PASS @ 944dfc064a572c2da5f8d73bb0ca7631c6741b91

PR #273, opened 2026-08-18
refactor(scan): give each target's probe timeout its own scope

### LINKED ISSUE #269: context leak: final per-target timeout context never canceled in 4 native probe modules

btuvay:
## Problem

`go vet` reports a possible context leak in four native probe modules (#269) — `ftp`, `mysql`, `smtp` and `winrm`, two findings each, eight in total, blocking a clean vet run.

**The report is real; its premise is not.** All four already release the last context, through a deferred closure that each `Execute` declares right after `var cancel context.CancelFunc`. What `lostcancel` cannot do is see a cancel that happens inside a deferred func literal, so it flags the `context.WithTimeout` assignment and every `return` below it.

Measured before changing anything: a test that captures the contexts the loop hands to the probe shows **both** of them, including the last, carrying `context canceled` by the time `Execute` returns — on `main`, before this change.

## User / Ops Impact

None. Nothing was leaking, and nothing about scanning changes. The value is a vet run whose findings are all worth reading, and a loop that is easier to be right about.

## Fix Summary

Give the timeout a scope instead of a lifetime that spans iterations.

The candidate keys are already sorted by target, so each target's candidates form one contiguous run. Handing that run to `probeTargetCandidates` lets the helper create the timeout and `defer cancel()` it three lines later — the shape `lostcancel` is built to recognise, and the shape the rest of this file already uses for single-shot probes.

The cross-iteration dance goes away with it: no `currentTarget` string, no cancel belonging to the previous target, no `cancel = nil` reset. Four files, the same edit in each, because the four share this loop verbatim.

## Behavior Change

None. Same candidates, same order, one timeout per target as before, same outputs on the same channel. The difference is where the release happens: at the end of the target's own work rather than at the start of the next target's, or in a deferred closure at function exit.

## Evidence

- `GOWORK=off go vet ./pkg/modules/scan/...` — **0 findings**, down from 8. That is the point of the change.
- `GOWORK=off go build ./...` and `go test ./pkg/modules/scan/...` — clean, full package green including every pre-existing probe test.
- **Two new tests** — `TestFTPNativeProbe_ReleasesEveryTargetTimeout` and `TestWINRMNativeProbe_ReleasesEveryTargetTimeout` — capture the per-target contexts through the module's own probe seam and assert each carries a deadline and is released before `Execute` returns. Two modules rather than one because the four share the loop verbatim and a property proved on one of them proves nothing about the rest.
- **Mutation-tested:** replacing `defer cancel()` with `_ = cancel` in both modules fails both tests; restored, both pass. The tests observe the property rather than the code shape.

## Risk / Rollback

Low. No behaviour change, no new dependency, no config. The one thing worth a reviewer's eye is the grouping loop — it assumes `keys` is sorted by target, which is what `sort.Slice` above it guarantees (target first, then port). Rollback is reverting this commit.

## Linked Issue

Closes #269

### DIFF
--- a/pkg/modules/scan/ftp_native_probe.go
+++ b/pkg/modules/scan/ftp_native_probe.go
@@ -238,29 +238,40 @@ func (m *ftpNativeProbeModule) Execute(ctx context.Context, inputs map[string]an
 		return left.target < right.target
 	})
 
-	currentTarget := ""
-	var targetCtx context.Context
-	var cancel context.CancelFunc
-	defer func() {
-		if cancel != nil {
-			cancel()
+	// keys are sorted by target first, so each target's candidates are one
+	// contiguous run. Handing the run to a helper gives the per-target timeout
+	// a scope to be created and released in, instead of one iteration canceling
+	// the context the previous iteration made (cyprob#269).
+	for start := 0; start < len(keys); {
+		target := candidates[keys[start]].target
+		end := start
+		for end < len(keys) && candidates[keys[end]].target == target {
+			end++
 		}
-	}()
+		m.probeTargetCandidates(ctx, keys[start:end], candidates, outputChan)
+		start = end
+	}
+
+	return nil
+}
+
+// probeTargetCandidates probes every candidate of one target under a single
+// timeout, and releases that timeout before it returns.
+func (m *ftpNativeProbeModule) probeTargetCandidates(
+	ctx context.Context,
+	keys []string,
+	candidates map[string]ftpProbeCandidate,
+	outputChan chan<- engine.ModuleOutput,
+) {
+	targetCtx := ctx
+	if m.options.TotalTimeout > 0 {
+		var cancel context.CancelFunc
+		targetCtx, cancel = context.WithTimeout(ctx, m.options.TotalTimeout)
+		defer cancel()
+	}
 
 	for _, key := range keys {
 		candidate := candidates[key]
-		if candidate.target != currentTarget {
-			if cancel != nil {
-				cancel()
-			}
-			currentTarget = candidate.target
-			targetCtx = ctx
-			cancel = nil
-			if m.options.TotalTimeout > 0 {
-				targetCtx, cancel = context.WithTimeout(ctx, m.options.TotalTimeout)
-			}
-		}
-
 		result := probeFTPDetailsFunc(targetCtx, candidate.target, candidate.hostname, candidate.port, candidate.protocolHint, m.options)
 		outputChan <- engine.ModuleOutput{
 			FromModuleName: m.meta.ID,
@@ -270,8 +281,6 @@ func (m *ftpNativeProbeModule) Execute(ctx context.Context, inputs map[string]an
 			Target:         candidate.target,
 		}
 	}
-
-	return nil
 }
 
 func defaultFTPProbeOptions() FTPProbeOptions {
--- a/pkg/modules/scan/mysql_native_probe.go
+++ b/pkg/modules/scan/mysql_native_probe.go
@@ -228,29 +228,40 @@ func (m *mysqlNativeProbeModule) Execute(ctx context.Context, inputs map[string]
 		return left.target < right.target
 	})
 
-	currentTarget := ""
-	var targetCtx context.Context
-	var cancel context.CancelFunc
-	defer func() {
-		if cancel != nil {
-			cancel()
+	// keys are sorted by target first, so each target's candidates are one
+	// contiguous run. Handing the run to a helper gives the per-target timeout
+	// a scope to be created and released in, instead of one iteration canceling
+	// the context the previous iteration made (cyprob#269).
+	for start := 0; start < len(keys); {
+		target := candidates[keys[start]].target
+		end := start
+		for end < len(keys) && candidates[keys[end]].target == target {
+			end++
 		}
-	}()
+		m.probeTargetCandidates(ctx, keys[start:end], candidates, outputChan)
+		start = end
+	}
+
+	return nil
+}
+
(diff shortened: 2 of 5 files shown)


cyprobot (approved):
Verified in a worktree at 450006bf.

**Correcting my own record first.** I filed `#269`, and its central technical claim — "the final context is never explicitly canceled" — is wrong. I missed the `defer func() { if cancel != nil { cancel() } }()` right after each file's `var cancel context.CancelFunc` declaration, which already releases the last context at function return. I wrote a standalone test against `origin/main` (before this PR touched anything) using the same probe-seam pattern this PR's own tests use, captured every context handed to `probeFTPDetailsFunc`, and checked `ctx.Err()` after `Execute` returned: both were already `context canceled`, including the last one. There was never a leak — `go vet`'s `lostcancel` just can't trace a cancel called through a nil-checked deferred closure over an outer-scope variable, which is exactly the shape #269 hit and this PR's own reasoning states. Apologies for the bad premise; worth being explicit about it since it's now a public issue.

**On its own merits, this PR is still worth taking.** It doesn't fix a bug (there wasn't one), but it removes a real false-positive from `go vet` and replaces a cross-iteration cancel dance with a shape that's genuinely easier to reason about — a scoped helper per target, matching the file's own single-shot-probe pattern.

**Verified the one thing the PR itself flags as worth a reviewer's eye**: the grouping loop assumes `keys` is sorted by target first. Checked all four files' `sort.Slice` comparators directly — target is the primary key, port priority only breaks ties within a target, in every one of the four. The assumption holds.

**Mutation-tested it**: replaced `defer cancel()` with `_ = cancel` in the FTP and WinRM helpers — both new tests (`TestFTPNativeProbe_ReleasesEveryTargetTimeout`, `TestWINRMNativeProbe_ReleasesEveryTargetTimeout`) fail exactly as the PR describes. Restored, clean.

**Added my own adversarial case the PR's tests don't cover**: one target with two candidate ports plus a second target with one, to check cross-candidate context-sharing within a target and non-interference across targets. Confirmed both of the first target's candidates receive the *same* context object (one timeout scope covering both), it's not yet canceled while either is still being probed, the second target gets a distinct context, and every context is released only after `Execute` returns as a whole. Removed the scratch test before finishing.

`go vet ./pkg/modules/scan/...` — 0 findings (down from 8, confirming `#269` closes clean). Full package green. `golangci-lint` clean. `gofmt` flags `winrm_native_probe.go` for pre-existing struct-tag misalignment, unrelated to this PR — confirmed by running `gofmt -l` against the same file on `origin/main` directly, same finding there. Not blocking, not introduced here. Real CI green across the board.

VERDICT: PASS @ 450006bf7455cb6b7bb94ce8599b25c23be7a1d3

PR #274, opened 2026-08-20
docs(test): validation report reproducibility + regex guard test (rebase of #195)

cyprobot:
## What this is

A rebase-and-revalidate of #195 (weissarc, opened 2025-11-05 off a fork
still named `vulntor`, predating at least two renames — the branch could
not be pushed to directly, so this is a fresh PR carrying the same content
plus one correction found along the way).

Original content, preserved as-is:
- Add a reproducibility block (commit hashes, commands) to
  `pkg/fingerprint/VALIDATION_REPORT.md`.
- Add `regex_guard_test.go`: a unit test guarding against YAML `\s`
  whitespace-pattern escaping regressions in HTTP `Server:` header
  matching.

## Why not just re-open #195 as-is

Atlas asked for the "non-functional, safe to merge" claim to actually be
checked against current main rather than trusted on a year-old green CI
run (that CI ran against a completely different `main`). It mostly held,
but not entirely — see below.

## What I found and fixed

- The rebase itself (`git rebase origin/main` on the PR's original commits)
  applied clean, no conflicts. `regex_guard_test.go`'s API usage
  (`StaticRule`, `NewRuleBasedResolver`, `Input`, `Resolve`) is unchanged on
  current main - the test compiles and passes as originally written.
- The reproducibility block was **not actually non-functional as written**:
  it named `27e5d35` (Phase 7, the PR's own base commit) as the "Rules
  commit"/"Dataset commit". But `VALIDATION_REPORT.md`'s own content — the
  file this block gets appended to — was rewritten twice after that, same
  day: `54ad03c` ("Phase 8 - Production-Ready Milestone", the real content
  change) and `6d17f21` (Pentora→Vulntor rebrand, text-only). The file
  currently reads "Report Version 2.0 (Phase 8...)" — a reproducibility
  pointer citing `27e5d35` would have pointed a reader at the wrong commit
  for the numbers actually sitting in the file. Fixed to cite `54ad03c`,
  with a one-line note on why (separate commit on top, not squashed into
  the original two, so the correction is visible on its own).

## Evidence

- `git rebase origin/main` on the PR's two original commits: clean, no
  conflicts.
- `go build ./...` — clean.
- `go vet ./...` — clean, whole module.
- `go test ./pkg/fingerprint/...` — green, including the new guard test
  (`TestRegexWhitespaceGuard_HTTPServerHeader`) run individually and as
  part of the full package.
- `golangci-lint run --new-from-rev=origin/main ./pkg/fingerprint/...` — 0
  issues.
- Actually ran the reproducibility block's own referenced commands rather
  than just checking they parse: `go test -v ./pkg/fingerprint -run
  TestValidationRunner` passes; `scripts/compare-validation-metrics.sh`
  exists at the referenced path.
- `gofmt -l` on the new test file — clean.

## Risk / Rollback

Two files, additive only (one new test, eight lines appended to a docs
file). No production code path touched. Revert is a straight `git revert`.

## Attribution

Original authorship (weissarc, #195) preserved in the first two commits;
the correction is a separate third commit on top so it's clear what
changed and why during rebase, rather than rewriting their work.

Closes #195


cyprobot (commented):
Verified in a worktree at ee6d2caf.

**The reproducibility fix is correct, and I checked it independently rather than trusting the commit hashes.** Confirmed via `git log`/`git show`: `27e5d35` (Phase 7, same-day, earlier) touches this file with content that predates the rewrite — `54ad03c` (Phase 8, same day, later) is the commit whose tail text (`Report Version 2.0 (Phase 8...)`, `2025-11-05 (Phase 8.3 completion)`) matches what's actually in the file today, modulo the later text-only Pentora→Vulntor rebrand (`6d17f21`). Citing `54ad03c` is right; `27e5d35` would have been wrong. Good catch, and disclosed as its own commit rather than folded into weissarc's original work — the right way to carry someone else's PR forward.

**The finding from my #195 review is unaddressed, and it's the substantive one.** I flagged that `TestRegexWhitespaceGuard_HTTPServerHeader` builds its own `StaticRule` in Go rather than reading `loadBuiltinRules()`/the shipped YAML, so it never actually exercises the YAML-escaping defect class it's named for — demonstrated at the time with `imap.generic`'s live double-escaping bug (since fixed as `#241`, unrelated to this PR). The test in this PR is byte-for-byte the one I looked at then — checked the file directly, not just the PR body's claim that "the test compiles and passes as originally written."

Re-ran the demonstration against *current* main rather than reusing the old one, since `#241` closing could have mooted it: temporarily reintroduced the exact double-escape regression into the shipped `http.nginx` rule (`server:\s*nginx` → `server:\\s*nginx`) and ran both the real resolver against `loadBuiltinRules()` and this PR's guard test side by side. The real resolver breaks — `Resolve` on `"Server: nginx"` returns `no matching rule found`. The guard test — same broken YAML on disk — still passes, untouched, because it never reads that file. Restored the YAML before finishing; nothing left mutated in the tree.

This means the test does what I said it does: it would stay green through a real regression of exactly the class in its own name. My suggested fix from the `#195` review still stands — assert over `loadBuiltinRules()`, either as a database-wide check (`\\` not appearing in any shipped `Match`, which is what actually would have caught `#241`) or by resolving a real banner through the shipped rules rather than a hand-built one.

Reproducibility block's own commands work as claimed: `go test -v ./pkg/fingerprint -run TestValidationRunner` passes, `scripts/compare-validation-metrics.sh` exists at the referenced path. Build/vet/test/lint all clean on the rest. CI green (still finishing `Build Binary (macos-latest)`/`Integration Tests` at review time, everything else already green).

Not blocking the reproducibility half — that's ready. The test file needs the same fix I asked for on `#195` before I'd pass this.

VERDICT: CHANGES @ ee6d2caf46a7e301b1200e11cf7eeb93aff5eade

cyprobot (commented):
Re-verified at 007ef1a.

Both fixes are exactly what I asked for, and I didn't just read them — mutation-tested both against the shipped catalog myself: reintroduced the identical double-escape regression into `http.nginx` (`server:\s*nginx` → `server:\\s*nginx`) and ran `TestRegexWhitespaceGuard_HTTPServerHeader` plus the new `TestNoShippedRuleHasDoubleEscapedMatch`. Both fail correctly — the first with `no matching rule found` against the shipped rule (not a hand-built stand-in), the second by naming `http.nginx` and printing the exact bad pattern. Restored the YAML, reran clean, both pass. `git status` clean after — nothing left mutated in the tree.

The database-wide test is the stronger of the two and does what I originally suggested: it would have caught `#241`'s `imap.generic` regression on its own, and it covers every rule shipped today plus every rule added later, not just `http.nginx`. No false positives against the real 50-rule catalog — checked the clean run above.

Full `pkg/fingerprint` package green, `go vet` clean, `golangci-lint --new-from-rev=origin/main` clean (0 issues — the stale-cache warning in my own local run pointed at `fingerprinter_registry.go`, a file this PR doesn't touch, same environment artifact seen on unrelated reviews before). Real CI fully green across all ten checks.

Reproducibility block from the previous pass still holds — nothing changed there.

Nothing further outstanding. Good turnaround.

VERDICT: PASS @ 007ef1ac901306e670acf7eef06d35b0bf4063c6

Issue #275, opened 2026-08-20
Lenovo IMM2 / sfcHttpd BMC interfaces have clear vendor strings, no fingerprint rule


cyprobot:
## Problem

`192.168.0.43` (office network) has a BMC-style management interface with two
clearly self-identifying banners, neither recognized:

```
5985/tcp, 5986/tcp (WinRM-labeled): Server: Lenovo IMM2 Web Server
5988/tcp, 5989/tcp (CIM/WBEM):      Server: sfcHttpd
```

`service_product`/`service_vendor` are empty for all four rows in
`asset_services`; `tech_tags` carries only protocol-level tags (`tls`,
`https`, `http_server`). Measured live, on a freshly re-scanned box running
current `main` (post cyprob#266) — not stale data.

Likely the same physical device: Lenovo's IMM2 (Integrated Management
Module 2, their BMC) commonly exposes both a WinRM-style management
endpoint and an SFCB (Small Footprint CIM Broker, `sfcHttpd`) CIM/WBEM
interface side by side — this looks like one device's two management ports,
not two unrelated services.

## Ask

Add fingerprint rules to `pkg/fingerprint/data/fingerprint_db.yaml`:
- Match `Server: Lenovo IMM2 Web Server` → vendor Lenovo, product IMM2 (BMC).
- Match `Server: sfcHttpd` → same vendor/device if confirmed to be IMM2's
  CIM interface, or its own generic "CIM/WBEM management interface" entry if
  it turns out to appear on non-Lenovo devices too — worth checking whether
  `sfcHttpd` is Lenovo-specific or a shared third-party component (SFCB is
  open source and used by multiple BMC vendors) before assuming it's always
  Lenovo.

Same shape as today's `libwebsockets`/RealVNC work (cyprob#265/#266) — read
the banner, don't guess from the port.

## Not established

Whether these two ports are actually the same physical BMC as the host's
other services (137/443 etc. also on .43) — reasonable inference from the
banners, not confirmed via a shared serial/MAC.

cyprobot:
Fixed by #277 (merged `d5f28231`) — added `http.lenovo_imm2` and `http.sfcb` fingerprint rules. Talos verified independently at both a1d06e08 and 7b3b285e (re-verified after validation_dataset.yaml entries were added), full suite green. Also confirmed live: the same sfcHttpd signature showed up on a DC host (10.20.29.249:5989) during today's real DC-network scan, not just the original office box — this rule covers both.

PR #277, opened 2026-08-20
feat(fingerprint): identify Lenovo IMM2 and SFCB BMC banners

cyprobot:
## Problem

`192.168.0.43` (office network) has a BMC-style management interface with
two clearly self-identifying `Server:` banners, neither recognized:

```
5985/tcp, 5986/tcp (WinRM-labeled): Server: Lenovo IMM2 Web Server
5988/tcp, 5989/tcp (CIM/WBEM):      Server: sfcHttpd
```

`service_product`/`service_vendor` were empty for all four rows in
`asset_services`; `tech_tags` carried only protocol-level tags.

## Fix

Two new rules in `pkg/fingerprint/data/fingerprint_db.yaml`:

- `http.lenovo_imm2` — matches `Server: Lenovo IMM2 Web Server` →
  vendor Lenovo, product Integrated Management Module II (IMM2).
- `http.sfcb` — matches `Server: sfcHttpd` → vendor SBLIM, product
  SFCB (Small Footprint CIM Broker).

On the "is this Lenovo-specific" question the issue raised: SFCB/sfcHttpd
is the HTTP front end of the open-source SBLIM CIM broker project, bundled
into multiple vendors' BMC firmware — not exclusive to Lenovo. Attributed
the rule to the upstream project (SBLIM), not the device vendor it happens
to ship inside of on this one host, per the issue's own suggested fallback.

## Evidence

Measured live against the reporting host, not inferred from the banner
strings alone:

- Grabbed raw `Server:` headers from all four ports directly (`curl -i`)
  before writing anything — confirmed the exact banner text.
- Ran a real targeted scan (`cyprob scan -t 192.168.0.43 -p
  5985,5986,5988,5989`) with the fix applied — `service.product`,
  `parsed_attributes.vendor`, and `parsed_attributes.cpe` are populated
  with `field_sources.vendor: "fingerprint"` for all four ports.
- A/B'd against a binary built from the same tree *without* the fix —
  before: `product` is a bare banner-string heuristic and `vendor` is
  empty (matches the issue's reported symptom exactly); after: both
  populated via the new fingerprint rules. Confirms the fix, not some
  other variable, is what moved the fields.
- `cyprob fingerprint validate --strict` — clean.
- `go test ./pkg/fingerprint/...` — passes, including the
  double-escape/regex-collision guard added in #274.

## Not established

Whether these four ports are the same physical BMC as the host's other
services (137/443 etc.) — reasonable inference from the banners, not
confirmed via a shared serial/MAC. Not addressed by this PR (not what
was asked; fingerprint identification doesn't need that confirmed).

## Risk / Rollback

Two new additive fingerprint rules, no existing rule modified. Worst case
if a match pattern is too broad: a wrong product/vendor label on some
other service's banner — bounded by the regex specificity ("lenovo imm2
web server" and "sfchttpd" are both full product-name strings, not generic
words) and by `go test ./pkg/fingerprint/...`'s collision guard, which
passes. Rollback is reverting this one file.

Refs #275

### DIFF
--- a/pkg/fingerprint/data/fingerprint_db.yaml
+++ b/pkg/fingerprint/data/fingerprint_db.yaml
@@ -781,6 +781,41 @@ rules:
     pattern_strength: 0.95
     port_bonuses: [902]
 
+  - id: 'http.lenovo_imm2'
+    protocol: 'http'
+    description: 'Lenovo IMM2 (Integrated Management Module II) BMC web server banner'
+    product: 'Integrated Management Module II (IMM2)'
+    vendor: 'Lenovo'
+    cpe: 'cpe:2.3:o:lenovo:integrated_management_module_ii_firmware:*:*:*:*:*:*:*:*'
+    match: 'server:\s*lenovo imm2 web server'
+
+    soft_exclude_patterns:
+      - "error"
+      - "unavailable"
+
+    pattern_strength: 0.95
+    port_bonuses: [5985, 5986]
+
+  # sfcHttpd is the HTTP front end of SFCB (Small Footprint CIM Broker, the
+  # SBLIM project) -- an open-source CIM/WBEM broker bundled into several
+  # vendors' BMC firmware (observed alongside Lenovo IMM2, but SFCB itself
+  # isn't Lenovo's). Attributed to the upstream project, not the device
+  # vendor it happens to ship inside on any one host.
+  - id: 'http.sfcb'
+    protocol: 'http'
+    description: 'SFCB (Small Footprint CIM Broker) CIM/WBEM HTTP interface banner'
+    product: 'SFCB (Small Footprint CIM Broker)'
+    vendor: 'SBLIM'
+    cpe: 'cpe:2.3:a:sblim:sfcb:*:*:*:*:*:*:*:*'
+    match: 'server:\s*sfchttpd'
+
+    soft_exclude_patterns:
+      - "error"
+      - "unavailable"
+
+    pattern_strength: 0.85
+    port_bonuses: [5988, 5989]
+
   - id: 'vnc.realvnc'
     protocol: 'vnc'
     description: 'RealVNC or VNC protocol banner'
--- a/pkg/fingerprint/testdata/validation_dataset.yaml
+++ b/pkg/fingerprint/testdata/validation_dataset.yaml
@@ -316,6 +316,22 @@ true_positives:
     expected_version: ''
     description: 'libwebsockets server on the conventional VNC port, previously unmatched (cyprob-ee#157)'
 
+  - protocol: http
+    port: 5985
+    banner: 'HTTP/1.1 401 Unauthorized\r\nWWW-Authenticate: Basic realm="avocent.com"\r\nServer: Lenovo IMM2 Web Server'
+    expected_product: 'Integrated Management Module II (IMM2)'
+    expected_vendor: 'Lenovo'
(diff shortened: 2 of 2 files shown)


cyprobot (commented):
Verified in a worktree at a1d06e08.

**Reproduced the resolution myself**, not just the winner — a script calling `resolver.Resolve` directly against `LoadRulesFromFile` on this branch: both real banners resolve correctly (`Lenovo IMM2` at conf 1.00 on 5985/5986, `SFCB`/`SBLIM` at conf 0.90 on 5988/5989). Confirmed the "before" state independently too, on `origin/main` directly with the same script: both banners return `no matching rule found` — matches the reported symptom (empty vendor) exactly, not inferred from the PR's own claim.

**Checked for the collision class this codebase has bitten itself on before** (`#266`'s WSDAPI/WinRM lesson): `http.lenovo_imm2` shares `port_bonuses: [5985, 5986]` with the existing `http.winrm` rule. Fed a genuine WinRM banner (`Microsoft-HTTPAPI/2.0` + `soap+xml` + `Negotiate`) through the resolver on those same ports — still resolves to WinRM/Microsoft, untouched, because the Lenovo rule's match string has no overlap with WinRM's signal. No rule elsewhere in the file references `5988`/`5989` at all. Also checked a banner on the wrong port (Lenovo signal on an arbitrary port) and an error-status SFCB response (exercises the `soft_exclude_patterns: ["error"]` penalty) — both behave as intended: content-only match without the port bonus, and a penalized-but-still-passing confidence rather than a false exclusion.

**One real gap, not blocking**: no `validation_dataset.yaml` entries for either new rule. This repo's own convention (164 dataset entries across 54 rules, roughly 3 per rule on average, and exactly what `#266` did for its own new rules) is what gives `TestRulePrecedence_NoTieDecidesAMatchInTheShippedCorpus` and the collision guards teeth against a *future* regression — right now, nothing in the automated suite would notice if someone later broke either of these two rules, only the live-host verification in this PR's own evidence section did, and that's not reproducible by CI. Worth a quick follow-up adding the two real banners as dataset entries, but the manual field verification here (real `curl`, real targeted scan, real A/B against an unpatched binary) is thorough enough that I'm not withholding the pass over it.

Correct call on attributing `http.sfcb` to SBLIM rather than Lenovo — matches the issue's own question and is verifiably accurate (SBLIM/SFCB is a real, vendor-independent open-source project).

`go build`, `go test ./pkg/fingerprint/...`, YAML parse all clean. Real CI green (a few `Build Binary`/`Integration Tests` jobs still finishing at review time, everything already reported is green).

VERDICT: PASS @ a1d06e08fa224734edb63a1545209305aae5f7ac

cyprobot (commented):
Re-verified at 7b3b285e.

Both new dataset entries use the real captured banners (including the "avocent.com" auth realm on the Lenovo one and the 501 response on the SFCB one — realistic field responses, not simplified). Ran `TestRulePrecedence_NoTieDecidesAMatchInTheShippedCorpus` explicitly — passes, so neither new entry introduced a tie against the rest of the 166-entry corpus. Full `TestValidationRunner` suite: 100% precision, matching the "0 false positives" claim independently rather than reusing the PR's own number. Full package green, YAML parses.

This closes the one gap from the last pass — both rules now have permanent regression coverage in the same corpus everything else in this file relies on.

VERDICT: PASS @ 7b3b285edf7b0f18e1fee56ddbb710068eba709f

Issue #276, opened 2026-08-20
MikroTik RouterOS HTTP title identifies the product; fingerprint_db.yaml doesn't extract it


cyprobot:
## Problem

`192.168.0.41` (MikroTik RouterOS device) is now correctly identified on two
ports after today's work — `21/tcp` ("RouterOS Telnet", vendor MikroTik) and
`22/tcp` ("ROSSSH") both show up right in `asset_services`. But `80/tcp` on
the same host, same device, carries only protocol-level tags
(`http`/`http_server`/`web_server`) — no product, no vendor — despite the
page itself stating the product outright:

```
<title>RouterOS router configuration page</title>
```

Measured live on a freshly re-scanned box running current `main`.

## Why this is worth fixing beyond this one host

This is the same shape as today's HP printer fix (title said "HP Neverstop
Laser MFP 1200w", now correctly tagged `hp_printer` across every HTTP port
on that device) — proof that title-based extraction works when it's wired
up. It just isn't wired up for this case. Any device that only exposes its
identity through its HTTP title (not a `Server:` header, not SNMP, not
mDNS) is likely hitting the same gap MikroTik's `80/tcp` hits here.

## Ask

Add a rule (or confirm/extend the existing title-matching mechanism used for
the HP printer case) so `<title>RouterOS router configuration page</title>`
resolves to vendor MikroTik / product RouterOS, consistent with what `21/tcp`
and `22/tcp` on the same host already report.

cyprobot:
Fixed by #278 (merged `c947a69f`) — added a fingerprint rule matching RouterOS's WebFig HTML title, bringing 80/tcp to the same identification level as 21/22 on the same device. Talos verified independently at d7f4d781 with his own adversarial cases (similar-but-unrelated titles, a fake error page, a migration-guide page mentioning RouterOS in passing) — none of them false-matched. Full CI green.

PR #278, opened 2026-08-20
feat(fingerprint): identify MikroTik RouterOS from its HTTP title

cyprobot:
## Problem

`192.168.0.41` (MikroTik RouterOS, office network) is correctly identified
on `21/tcp` ("RouterOS Telnet", vendor MikroTik) and `22/tcp` ("ROSSSH")
via their own native probes. `80/tcp` on the same device carries only
protocol-level tags (`http`/`http_server`/`web_server`) -- no product, no
vendor -- despite the page stating the product outright:

```
<title>RouterOS router configuration page</title>
```

There is no `Server:` header on this endpoint at all, only the title.

## Fix

New rule in `pkg/fingerprint/data/fingerprint_db.yaml`:

- `http.mikrotik_routeros` -- matches
  `<title>RouterOS router configuration page</title>` in the response
  body. This is the same matching mechanism every other rule in the file
  already uses (the resolver matches the full lowercased response, not
  just headers) -- see `http.lenovo_imm2` (cyprob#275) for the same shape
  matched against a header instead of a body string. No new mechanism
  needed, just a rule that targets the body.

Also added a `validation_dataset.yaml` entry alongside it (Talos flagged
this as a routine addition in #277's review, not just an on-request one).

## Evidence

Measured live against the reporting host, not inferred:

- `curl -i` against `192.168.0.41:80` directly, before writing anything --
  confirmed there's genuinely no `Server:` header, only the title in the
  body.
- Ran a real targeted scan (`cyprob scan -t 192.168.0.41 -p 21,22,80`)
  with the fix -- port 80's `service.product`/`parsed_attributes.vendor`/
  `parsed_attributes.cpe` are now populated with
  `field_sources.vendor: "fingerprint"`, matching what 21/22 already
  report via their native probes. 21/22 unaffected, as expected -- this
  rule only had to cover the gap on 80.
- A/B'd against a binary built from the same tree *without* the fix --
  before: port 80's product is a bare `"HTTP"` heuristic guess, vendor
  empty; after: both populated. Matches the issue's reported symptom
  exactly.
- `cyprob fingerprint validate --strict` -- clean.
- `go test ./pkg/fingerprint/...` -- passes, 0 false positives on the full
  167-case validation suite (up from 166 after #277, +1 for this rule).

## Not established

Whether MikroTik's WebFig title string is stable across RouterOS firmware
versions -- only measured on this one live device's current version. The
match targets the exact observed string rather than a looser pattern,
consistent with not guessing at variations that weren't measured.

## Risk / Rollback

One new additive fingerprint rule, no existing rule modified. The title
string is specific enough (`routeros router configuration page`) that a
collision with an unrelated device is unlikely, and `go test
./pkg/fingerprint/...`'s collision guard passes. Rollback is reverting
this one file.

Refs #276

### DIFF
--- a/pkg/fingerprint/data/fingerprint_db.yaml
+++ b/pkg/fingerprint/data/fingerprint_db.yaml
@@ -175,6 +175,25 @@ rules:
 
     pattern_strength: 0.85
 
+  # No Server: header on MikroTik's WebFig login page -- the only
+  # self-identifying string is the HTML <title>. Matched against the full
+  # response body like any other rule here (see http.lenovo_imm2 for the
+  # same shape against a header instead), so no new mechanism is needed.
+  - id: 'http.mikrotik_routeros'
+    protocol: 'http'
+    description: 'MikroTik RouterOS WebFig HTML title, no Server: header present'
+    product: 'RouterOS'
+    vendor: 'MikroTik'
+    cpe: 'cpe:2.3:o:mikrotik:routeros:*:*:*:*:*:*:*:*'
+    match: '<title>\s*routeros router configuration page\s*</title>'
+
+    soft_exclude_patterns:
+      - "error"
+      - "unavailable"
+
+    pattern_strength: 0.9
+    port_bonuses: [80, 443]
+
   # FTP Servers
   - id: 'ftp.pureftpd'
     protocol: 'ftp'
--- a/pkg/fingerprint/testdata/validation_dataset.yaml
+++ b/pkg/fingerprint/testdata/validation_dataset.yaml
@@ -332,6 +332,14 @@ true_positives:
     expected_version: ''
     description: 'sfcHttpd CIM/WBEM interface, attributed to the SBLIM project not the device vendor, previously unmatched (cyprob#275)'
 
+  - protocol: http
+    port: 80
+    banner: '<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd"><html><head><title>RouterOS router configuration page</title></head></html>'
+    expected_product: 'RouterOS'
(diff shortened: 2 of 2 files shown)


cyprobot (commented):
Verified in a worktree at d7f4d781.

Reproduced the resolution myself, plus wrote my own adversarial cases beyond what the PR/validation dataset cover:

- Real MikroTik banner on 80 and 443 — both resolve to `RouterOS`/`MikroTik` at conf 0.95, matching the claim.
- A generic, unrelated "Router Configuration Page" title (deliberately similar phrasing, different vendor) — correctly does **not** match. Confirms the pattern is anchored on the distinctive combination, not a generic phrase that could catch other embedded-device admin pages.
- A page containing the word "error" elsewhere in the body (an inline CSS class, not an actual error state) — still resolves, at reduced confidence (0.75) from the soft-exclude penalty, not a false negative. The rule's base strength (0.9) plus port bonus comfortably absorbs that penalty.
- A page that only mentions "RouterOS" in passing (a migration-guide title, not an actual WebFig page) — correctly does not match, since the exact `<title>...routeros router configuration page...</title>` phrase isn't present.

Ran `TestRulePrecedence_NoTieDecidesAMatchInTheShippedCorpus` and the full `TestValidationRunner` suite explicitly — both pass, 0 false positives on the (now 168-entry) corpus, independently confirming the "0 false positives" claim rather than reusing the PR's own number. Full package green, YAML parses on both files.

Good on incorporating the `validation_dataset.yaml` addition proactively this time rather than waiting to be asked again — this rule now has the same permanent regression coverage as everything else in the file.

The "not established" section (firmware-version stability of the title string) is honestly scoped and not something I'd expect resolved before merging a rule this narrowly targeted.

Real CI green on everything reported so far (`Build Binary`/one `Integration Tests` job still finishing at review time).

VERDICT: PASS @ d7f4d7810be309f2544fb658dee8c102b725002a

PR #282, opened 2026-08-21
fix: pace port scan probes and stop retrying refusals

cyprobot:
## Summary

Three surgical fixes to `pkg/modules/discovery/tcp_port_discovery.go`'s load profile against a single host, closing the three priority items from cyprob-ee#97:

1. **Stop retrying refused connections.** `dialWithRetries` broke only on `err == nil`, so a closed port that RSTs instantly was still dialed `Retries+1` times (4x at the forced `top_1000` `retries=3`) - a definitive answer retried for zero new information.
2. **Stop treating EHOSTUNREACH as a transient error.** It was falling into the `otherErrors` bucket and getting re-probed by the verification pass at the *longer* timeout - the opposite of correct behavior, since EHOSTUNREACH is the target's own back-off signal (field-measured on this estate: 30 rapid connects to a management controller produced it on 443/8443). It's still reported under `OtherErrorPorts` in the result (no schema change), just no longer retried or reverified.
3. **Wire up the pacing EE already ships.** `batch_enabled`/`batch_size`/`batch_delay` have been built and plumbed by EE (`adapter.go`) since before this module read any of them - `grep`ping those three names across all of CE's `pkg/` returned zero hits before this PR. When enabled, a target's ports are now probed in `batch_size` groups with `batch_delay` between groups, instead of `Concurrency` alone bounding how many connections a single host receives at once.

Default behavior is unchanged: `batch_enabled` defaults to `false`, so nothing paces unless EE (or a caller) turns it on.

## Evidence

- **Field-reproduced the root cause independently**, against the exact host cyprob-ee#97's comment named (`192.168.0.62`): 8 back-to-back `top_1000`-shaped scans at `concurrency=100`, no batching - port 80 came back closed on 3/8 runs, alternating with no other variable changed. Same shape, same order of magnitude as the issue's own field measurement (3/5 closed). Confirms the concurrency-storm mechanism before fixing it, not just arithmetic from the code.
- Unit tests added for all three fixes (`dialWithRetries` stops at attempt 1 for both `ECONNREFUSED` and `EHOSTUNREACH`; EHOSTUNREACH ports are excluded from `reverifyPorts()` while still landing in `OtherErrorPorts`; `batch_enabled`/`batch_size`/`batch_delay` parse correctly from `Init`; `scanPortBatch` measurably paces when enabled and is unaffected when disabled).
- `go build ./...`, `go vet ./pkg/modules/discovery/...`, `golangci-lint run ./pkg/modules/discovery/...` all clean.
- All pre-existing tests in the package still pass unmodified.

## Test plan

- [x] `go test ./pkg/modules/discovery/...` - all pass, including new coverage
- [x] `golangci-lint run ./pkg/modules/discovery/...` - 0 issues
- [x] Live A/B against the field-confirmed flaky host (192.168.0.62) - baseline reproduced the reported flake

Closes cyprob-ee#97 items 1-3 (item 4, the TLS evidence probe / SMB probe scope, is explicitly lower priority and out of scope here).

### DIFF
--- a/pkg/modules/discovery/tcp_port_discovery.go
+++ b/pkg/modules/discovery/tcp_port_discovery.go
@@ -43,6 +43,16 @@ type TCPPortDiscoveryConfig struct {
 	Retries                 int                   `json:"retries"`
 	StopOnFirstOpen         bool                  `json:"stop_on_first_open"`
 	VerificationPassEnabled bool                  `json:"verification_pass_enabled"`
+	// BatchEnabled paces a target's port probes in groups of BatchSize,
+	// waiting BatchDelay between groups, instead of letting Concurrency alone
+	// decide how many connections a single host receives at once. EE has
+	// plumbed batch_enabled/batch_size/batch_delay through since before this
+	// module read them (cyprob-ee#97) - a single scan of one host could open
+	// up to Concurrency simultaneous connections against it with nothing to
+	// space them out.
+	BatchEnabled bool          `json:"batch_enabled,omitempty"`
+	BatchSize    int           `json:"batch_size,omitempty"`
+	BatchDelay   time.Duration `json:"batch_delay,omitempty"`
 }
 
 // TCPPortDiscoveryModule implements the engine.Module interface for TCP port discovery.
@@ -191,6 +201,24 @@ func newTCPPortDiscoveryModule() *TCPPortDiscoveryModule {
 					Required:    false,
 					Default:     defaultConfig.VerificationPassEnabled,
 				},
+				"batch_enabled": {
+					Description: "Pace a target's port probes in groups of batch_size, waiting batch_delay between groups, instead of sending all of them at once (bounded only by concurrency).",
+					Type:        "bool",
+					Required:    false,
+					Default:     defaultConfig.BatchEnabled,
+				},
+				"batch_size": {
+					Description: "Number of ports probed per group when batch_enabled is true.",
+					Type:        "int",
+					Required:    false,
+					Default:     0,
+				},
+				"batch_delay": {
+					Description: "Pause between port groups when batch_enabled is true (e.g., '500ms').",
+					Type:        "duration",
+					Required:    false,
+					Default:     "0s",
+				},
 			},
 			// ActivationTriggers: Usually none for a primary discovery module, unless it depends on a very specific prior state.
 			// IsDynamic: false,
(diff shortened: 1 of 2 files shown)


cyprobot (commented):
Verified in a worktree at a817f8e3.

**Traced all three fixes end to end, not just the changed lines.** `dialWithRetries`'s new `break` sits correctly inside the retry loop, falls through to the existing `return nil, lastErr` — confirmed by reading the surrounding function, not just the diff hunk. `reverifyPorts()` reads only `timedOut`/`otherErrors`, confirmed `backoff` (EHOSTUNREACH) is genuinely excluded by reading its full body, and `otherErrorPorts()` correctly merges `backoff` back in for external-schema purposes without also feeding it to `reverifyPorts()` — the two accessors diverge exactly where the fix needs them to. `scanPortBatch`'s caller (`scanTargetPortsAll`) applies pacing to both the first pass and the reverify pass, which is right — a paced reverify pass is still a reverify pass hitting the same host.

**Mutation-tested all three independently, not just re-run the PR's own tests:**
- Removed the `isRefusedLikeError(err) || isHostUnreachableError(err)` break — `TestTCPPortDiscoveryModule_DialWithRetriesDoesNotRetryRefusedConnection`/`...HostUnreachable` both fail, reporting exactly 4 attempts (`Retries+1` at the forced `top_1000` value of 3) instead of 1 — the precise quadrupling the issue measured.
- Removed the `isHostUnreachableError` case from `recordNegative`'s switch (so EHOSTUNREACH falls into the `default`/`otherErrors` bucket, the pre-fix behavior) — `TestTCPPortDiscoveryModule_ScanTargetPortsAll_HostUnreachableIsNotReverified` fails, dial count 2 instead of 1: the exact "re-probe the device that told us to back off" defect.
- Forced `chunkPortsForPacing` to always return one chunk (batching silently disabled regardless of config) — `TestTCPPortDiscoveryModule_ScanPortBatch_PacesWhenBatchEnabled` fails, elapsed time 187µs instead of ≥60ms.

All three restored cleanly afterward; full package green again.

**Checked the two guard clauses in `chunkPortsForPacing`** (`BatchSize <= 0`, `BatchSize >= len(ports)`) — both correctly collapse to the single-chunk (unpaced) path rather than producing a degenerate empty-chunk or single-port-per-chunk result, and `Init`'s own negative-`batch_size` warning-and-clamp-to-0 means a misconfigured negative value can't reach `chunkPortsForPacing` in a broken state either.

Full `go build`, `go test ./pkg/modules/discovery/...`, `go vet`, `golangci-lint`, `gofmt` all clean. Real CI green across the board.

The field-reproduction methodology (8 unpaced runs against the same host the issue named, matching the flake shape before touching any code) is exactly the right way to confirm the mechanism first — can't verify that specific network reproduction from here, but the code-level fix and its tests stand on their own regardless.

Correctly scoped: items 1-3 addressed, the two lower-priority items from the issue (TLS probe port scope, SMB probe deadline) correctly left out as unrelated to port-scan pacing/retry logic.

VERDICT: PASS @ a817f8e3692d63fe1f04d5b67f72a4b845902006

Issue #199, opened 2025-11-06
JSON output should include evidence field to show active probe results


jonaserflow:
## Problem

Users cannot see that active probing is working - only passive banners are visible in JSON output.

**Current Behavior** (Bad UX):
```json
{
  "service": {
    "name": "pop3",
    "raw_banner": "+OK Hello there.",
    "port": 110
  }
}
```
→ User only sees passive banner, doesn't know `pop3-capa` probe was executed

**Expected Behavior** (Better UX):
```json
{
  "service": {
    "name": "pop3",
    "raw_banner": "+OK Hello there.",
    "evidence": [
      {
        "probe_id": "tcp-passive",
        "response": "+OK Hello there."
      },
      {
        "probe_id": "pop3-capa",
        "response": "+OK\r\nCAPA\r\nTOP\r\nUIDL\r\nRESP-CODES\r\nPIPELINING\r\nAUTH-RESP-CODE\r\nSTLS\r\nSASL\r\n."
      }
    ],
    "port": 110
  }
}
```
→ User can see all probe results, verify Phase 1.5 is working

## Root Cause

- `scan.BannerGrabResult` struct has `Evidence` field that is populated
- However, during JSON serialization, this field is NOT included in output
- Only `raw_banner` field is shown to users

## Impact

- **User Experience**: Phase 1.5 (Probe Fallback) works successfully but users can't see it, so they think it's not working
- **Debugging**: Seeing which probes were executed is critical for troubleshooting
- **Transparency**: Active probing behavior should be transparent to users

## Proposed Solution

1. **Modify `scan.BannerGrabResult` JSON tags** (pkg/modules/scan/banner_grab.go):
   ```go
   type BannerGrabResult struct {
       IP       string     `json:"ip"`
       Port     int        `json:"port"`
       Protocol string     `json:"protocol"`
       Banner   string     `json:"banner"`
       Evidence []Evidence `json:"evidence,omitempty"` // ✅ Add JSON tag
       Timeout  bool       `json:"timeout"`
       Error    string     `json:"error,omitempty"`
   }
   
   type Evidence struct {
       ProbeID  string `json:"probe_id"`
       Response string `json:"response"`
       Protocol string `json:"protocol,omitempty"`
   }
   ```

2. **Update CLI output formatter** to include evidence in scan results

3. **Add `--show-evidence` flag** (optional) to reduce verbosity for users who don't need it

## Testing Requirements

- ✅ JSON output includes evidence array
[...]

PR #198, opened 2025-11-05
feat(fingerprint): implement Phase 1 and Phase 1.5 protocol detection fallback

jonaserflow:
## Summary

This PR implements **Phase 1** and **Phase 1.5** of the protocol detection system, enabling Pentora to detect services running on non-standard ports while maintaining production-quality validation metrics.

### Phase 1: Port-Based Plugin Lookup with Fallback

When the protocol hint is generic (tcp/udp) or unknown, the fingerprint resolver now tries ALL fingerprint rules as a fallback mechanism. This enables detection of services on unusual ports without requiring port-specific configuration.

**Implementation:**
- Modified `pkg/fingerprint/resolver_rulebased.go` to add `useFallback` logic (lines 84-95)
- Modified `pkg/modules/parse/fingerprint_parser.go` to trigger fallback by setting `protocolHint` to empty string (lines 142-151)

**How it works:**
1. If protocol hint is specific (e.g., "http", "ssh"), use fast path (protocol-specific rules only)
2. If protocol hint is generic ("tcp", "udp", ""), activate fallback mode (try all rules)
3. Banner matching still uses confidence scoring and exclusion patterns

### Phase 1.5: Probe Fallback System (3-Tier)

When no port-specific probes are found AND passive banner grabbing yields no results, the scanner now tries a curated set of fallback probes commonly found on non-standard ports.

**Implementation:**
- Added `FallbackProbeIDs` field to `ProbeCatalog` struct
- Added `FallbackProbes()` method to return fallback probes (lines 67-95)
- Modified `pkg/fingerprint/data/probes.yaml` to define fallback probes (lines 1-9)
- Modified `pkg/modules/scan/banner_grab.go` to use fallback probes (lines 317-331)

**3-Tier Probe Selection:**
1. **Port-Based**: If port has known protocol hints, use port-specific probes
2. **Fallback**: If no port-specific probes match AND passive banner is empty, try fallback probes
3. **Give Up**: If both tiers fail, skip active probing

**Fallback Probes:**
- `http-get`: Most common (web services on unusual ports like cPanel, proxies)
- `https-get`: HTTPS variant for TLS services
- `redis-ping`: Database fallback (Redis, Memcached on custom ports)
- `ftp-feat`: FTP fallback for file transfer services

## Real-World Use Cases Enabled

This implementation solves detection problems for:

- **cPanel web hosting**: HTTP on ports 2082, 2083, 2086, 2087, 2095, 2096
[...]

PR #283, opened 2026-08-21
fix: replace batching's chunk barrier with a sliding-window worker pool

cyprobot:
## Summary

Follow-up to cyprob/cyprob#282, addressing what cyprob-ee#198's live testing found: the previous pacing implementation split a target's ports into fixed-size groups and waited for the entire group - including its slowest straggler - to finish before starting the next group. Deployed live (cyprob-ee#199, reverted), this cost far more than `batch_size x batch_delay` predicts: one filtered port (a full per-attempt timeout) inflated the cost of every other port sharing its group, and because groups ran strictly sequentially, that cost compounded across every group after it. Isolated single-host measurement: 16.5s unpaced -> 32.5s paced. A full-subnet scan's total runtime grew enough to collide with an unrelated 10-minute execution timeout, silently dropping a host that hadn't been reached yet.

**Fix:** replaced the chunk barrier with a fixed-size worker pool. `batch_size` workers pull ports from a shared queue; each pauses `batch_delay` after finishing one port before picking up its next. A slow port only blocks the one worker that drew it - the other `batch_size - 1` workers keep draining the queue independently. Each target still gets a real, tight concurrency ceiling of `batch_size` (separate from the global `Concurrency` semaphore, which still bounds the total across every target in a multi-host scan).

**Also fixed:** a latent data race `-race` caught while testing this. `tcpPortScanOutcomes`'s read accessors (`openPorts`/`timedOutPorts`/`refusedPorts`/`otherErrorPorts`) never took the struct's own mutex - safe before only because every call site read after a `wg.Wait()` happens-before barrier. This PR's own new test needed a genuine concurrent read (checking progress mid-scan), which correctly surfaced the gap. All four accessors now lock, matching the writers.

## Evidence

- New regression test (`TestTCPPortDiscoveryModule_ScanPortBatch_SlowPortDoesNotBlockOtherWorkers`) reproduces the exact defect shape - one slow port among several fast ones - and asserts the other worker keeps making independent progress while the slow dial is still in flight. This would fail against the old chunk-barrier code (a slow port anywhere in a group blocks the whole group).
- Full package suite green under `-race`, including the new test run 5x in a loop for flakiness with no failures.
- `golangci-lint run` - 0 issues.
- **Live-verified against the same host cyprob-ee#198/#199 used (`192.168.0.62`):** the new paced path completes and correctly reports port 80 open (the baseline unpaced run in the same session flaked to closed). Reporting this honestly rather than only the favorable framing: pacing is still substantially slower than unpaced on this host once `retries=3` (the forced `top_1000` policy) amplifies genuinely-filtered ports' per-port dial cost - 12s unpaced vs 37s paced with `batch_size=25`/`batch_delay=100ms` on a 150-port range. That cost is inherent to trading concurrency for accuracy and is separate from the chunk-barrier defect this PR removes; it is not re-tuned here.

## Scope

This PR fixes the CE-side mechanism only. It does **not** re-enable `batch_enabled` by default in EE - cyprob-ee#199 was reverted for the reasons above and stays reverted. Whether/when to re-enable a default, and what `batch_size`/`batch_delay`/`retries` values are actually worth the slowdown, is cyprob-ee#198's open question and needs its own live before/after evidence per that issue's guidance, not assumed here.

### DIFF
--- a/pkg/modules/discovery/tcp_port_discovery.go
+++ b/pkg/modules/discovery/tcp_port_discovery.go
@@ -736,11 +736,19 @@ func (m *TCPPortDiscoveryModule) scanTargetPortsAll(
 	return openPorts
 }
 
-// scanPortBatch probes ports, optionally paced in groups of m.config.BatchSize
-// with m.config.BatchDelay between groups (cyprob-ee#97): without pacing,
-// Concurrency alone bounds how many connections a single host receives at
-// once, and a single-target scan can open up to Concurrency of them
-// simultaneously against that one IP.
+// scanPortBatch probes ports, optionally paced to at most m.config.BatchSize
+// of them in flight for this target at once, with m.config.BatchDelay between
+// a worker finishing one port and starting its next (cyprob-ee#97/#198):
+// without pacing, Concurrency alone bounds how many connections a single host
+// receives at once, and a single-target scan can open up to Concurrency of
+// them simultaneously against that one IP.
+//
+// Pacing is a fixed-size worker pool (batch_size workers), not a sequence of
+// hard-barrier groups: an earlier design waited for every port in a group to
+// finish - including its slowest straggler - before starting the next group,
+// which let one filtered port (a full timeout) inflate the wall-clock cost
+// of an entire group. Here, a slow port only holds up its own worker; the
+// other batch_size-1 workers keep pulling from the shared queue.
 func (m *TCPPortDiscoveryModule) scanPortBatch(
 	ctx context.Context,
 	ip string,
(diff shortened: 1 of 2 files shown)


cyprobot (commented):
Verified in a worktree at 58d1773f.

**Noting my own gap first.** I passed `#282` after mutation-testing its chunk-barrier pacing, but every adversarial case I wrote used a fast synthetic dial (`net.Pipe()`), never a genuinely slow one — so the barrier's actual failure mode (one filtered port's full timeout compounding across every subsequent group) never showed up in anything I tested. The live appliance test that caught it did what my code-level review didn't. Reviewing this fix with that lesson applied.

**Confirmed the new test catches the exact old defect, on the old code, in two independent ways.** Checked out `#282`'s code (`a817f8e3`) in place of this PR's, reran `TestTCPPortDiscoveryModule_ScanPortBatch_SlowPortDoesNotBlockOtherWorkers` unmodified: without `-race`, it fails on the logic assertion alone (`"1" is not >= "3"` — only 1 port progressed at the 200ms checkpoint instead of 3+, the head-of-line block itself). With `-race`, it *also* trips a genuine data race on the very same run — meaning this one test independently surfaces both defects Helios describes fixing, not just the one it was written to target.

**Traced the worker-pool design for correctness**: `batch_size` workers `range` over a shared unbuffered `portCh`; a slow `dialOnePort` call only holds up the one worker that drew that port, the other workers keep pulling independently. The global `sem` semaphore is still acquired inside `dialOnePort`, so total concurrent dials across every target in a multi-host scan stay bounded exactly as before — this PR only changes how one target's own ports are paced, not the cross-target ceiling.

**Added three of my own adversarial cases, not just re-running the PR's:**
- Two slow ports (300ms each) scattered among 98 fast ones at `batch_size=3`: total elapsed 431ms, not the ~5-10s a compounding barrier would produce at that ratio — confirms no compounding, independent of the PR's own single-slow-port test.
- Context cancellation 30ms into a 200-port scan with a 500ms `BatchDelay`: `scanPortBatch` returns within 2s (actually near-immediately) — no goroutine hang from the worker pool's cancellation paths.
- Five goroutines hammering all four accessor methods concurrently through an entire scan, not just the PR's single mid-scan checkpoint read — clean under `-race`.

All three pass, all three ran under `-race`.

Full package green under `-race` (including the PR's own test run 5x via `-count=5`, and the whole package, not just the new test). `go vet`, `golangci-lint`, `gofmt` all clean.

**On the honestly-reported slowdown** (12s → 37s on the same live host): agree this is a separate, inherent cost-of-pacing question, correctly not conflated with the barrier-compounding bug this PR actually fixes, and correctly left to `#198`'s own re-tuning work rather than addressed here.

Correctly scoped: `EE`'s `batch_enabled` default stays off (`#199` stays reverted) — this is a CE-mechanism fix only, re-enabling by default is a separate decision with its own evidence bar.

Real CI green on everything reported so far.

VERDICT: PASS @ 58d1773f841d5073819da55eeee652fe75c86046

